PR #12, opened 2024-04-03
Arbitration node

emilesavoie:
TEST

With Target Arbitration = 0
Always printing 0 regardless of Heartbeat

With Target Arbitration = 1 (Teleop)
Printing constant 0 msg at no heartbeat
Printing constant teleop msg with working heartbeat


emilesavoie:
TEST

With Target Arbitration = 0
Always printing 0 regardless of Heartbeat

With Target Arbitration = 1 (Teleop)
Printing constant 0 msg at no heartbeat
Printing constant teleop msg with working heartbeat

PR #24, opened 2024-05-09
Science control + fix arb for teleop and can

emilesavoie:
MODIFS
changed sub in can to /drive_train/out/motors
rm forced arbitration

ADD-ONS
science controls


Pmichaud29 (approved):
mvp

PR #53, opened 2024-07-31
[dev/fix_abtr] Fix drive_train arbitration

Pmichaud29:
Working with GUI but toggle in gui doesn't follow actual arbitration, issue as been created to fix this later accordingly

### DIFF
--- a/rover_helper/launch/rover.launch.py
+++ b/rover_helper/launch/rover.launch.py
@@ -14,8 +14,8 @@ def generate_launch_description():
 
     return LaunchDescription([
         simulate_arm_arg,
-        IncludeLaunchDescription(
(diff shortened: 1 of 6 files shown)


rover_helper/launch/rover.launch.py
emilesavoie, line 1: Not sure this comment is supposed to stay unless we dont want can error messages printed to the console (might be the case)

emilesavoie (approved):
Check comment on rover_helper launch file. Does not affect functionnality

rover_helper/launch/rover.launch.py
Pmichaud29: true

PR #54, opened 2024-08-03
Cartesian mode

emilesavoie:
This implementation covers the cartesian control mode for the robotic arm and its simulation.

Note that values WILL have to be modified in order to fit real arm's behavior. The node currently only takes in simulation feedback and should be tested carefully.

Node has been tested and is fully functional


Pmichaud29:
Needed to be rebased

PR #64, opened 2024-08-10
RTSP stream redistribution

emilesavoie:
NOTES:
cams are smoooooth
TODO:
update r1m_ips.txt in ui to wanted cams in UI


Pmichaud29:
this won't be merged as it will be redesigned this year, branch type as been changed from dev to idle

PR #80, opened 2024-10-17
added macro to get path of package by name

### LINKED ISSUE #78: macro to get package path by name

Zach3292:
Added macro to get package path by name.

https://github.com/robotique-udes/rover/blob/a203cdfebf3f059c48698cffd9496915189956f5/rover_msgs/include/rovus_lib/macros.h#L53-L54

### DIFF
--- a/rover_msgs/include/rovus_lib/macros.h
+++ b/rover_msgs/include/rovus_lib/macros.h
@@ -49,5 +49,8 @@
 
 // Removes unused argument warning 
 #define REMOVE_UNUSED(x) (void)(x)
+
+#define GET_PACKAGE_SOURCE_DIR(package_name) \
+    (ament_index_cpp::get_package_prefix(package_name) + "/../../src/rover/" + package_name)
     
 #endif //__MACROS_H__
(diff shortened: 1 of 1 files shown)


rover_msgs/include/rovus_lib/macros.h
```diff
@@ -49,5 +49,8 @@
 
 // Removes unused argument warning 
 #define REMOVE_UNUSED(x) (void)(x)
+
+#define GET_PACKAGE_SOURCE_DIR(package_name) \
+    (ament_index_cpp::get_package_prefix(package_name) + "/../../src/rover/" + package_name)
```
Pmichaud29, line 56: Consider adding the necessary ament_index_cpp #include at the top of the file. Otherwise people using the macro will need to include it in their projects which defeats the purpose

Pmichaud29 (changes requested):

rover_msgs/include/rovus_lib/macros.h
```diff
@@ -1,3 +1,5 @@
+#include <ament_index_cpp/get_package_prefix.hpp>
+
```
Pmichaud29: Must be after #ifndef to speed up compile time

Pmichaud29 (changes requested):

rover_msgs/include/rovus_lib/macros.h
Zach3292: it has been moved

Pmichaud29 (approved):
Nice thanks!

PR #95, opened 2024-12-19
python api for ipcamera parameter change

ChristopherLajoie:


rover_auxiliary/src/ipcamera_api.py
```diff
@@ -0,0 +1,322 @@
+import logging, requests, signal, string
```
Pmichaud29: I would change the logging to use the default ROS logger, maybe take one as argument at init

rover_auxiliary/src/ipcamera_api.py
```diff
@@ -0,0 +1,322 @@
...
+        self.username = username
+        self.password = password
+
+        # Configure logging
+        for handler in logging.root.handlers[:]:
+            logging.root.removeHandler(handler)
+        logging.basicConfig(level=logging.INFO, format='%(levelname)s - %(message)s')
+        self.logger = logging.getLogger()
+
+        def timeout_handler(signum, frame):
```
Pmichaud29: Unless there's a specific use, I would define all function in the class at the same level (not inlined in the constructor). a lambda in the signal call would also work

rover_auxiliary/src/ipcamera_api.py
```diff
@@ -0,0 +1,322 @@
...
+            self.source_token = self.profile.VideoSourceConfiguration.SourceToken
+
+        except TimeoutException:
+            self.logger.error("Camera initialization timed out")
+        except Exception as e:
+            self.logger.error(f"Failed to initialize the camera: {str(e).split(':')[-1].strip()}")
+        finally:
+            signal.alarm(0)  
+
+    def set_onvif_param(self, param: str, value: str, extra: tuple=None):
```
Pmichaud29: Can this be declared private? Since there's no checks on params str or values, it would make sure a user of the lib doesn't call this in the wrong way.

rover_auxiliary/src/ipcamera_api.py
```diff
@@ -0,0 +1,322 @@
...
+                self.logger.info(f"{param} updated to {value}")
+            else:
+                self.logger.error(f"Failed to update {param} via HTTP. Status: {response.status_code}")
+        except Exception as e:
+            self.logger.error(f"Failed to set parameter {param}: {e}")
+
+    # ======================================================================================================================= #
+    #                                                      API METHODS                                                        #
+    # ======================================================================================================================= #
+
```
Pmichaud29, line 192: Can you add checks for all arguments/input values to make sure they are valid for our cams?

rover_auxiliary/src/ipcamera_api.py
```diff
@@ -0,0 +1,322 @@
...
+        self.set_http_param('LensDistortionCorrection', '0')
+    
+    def setAntiFog(self, value: int = 50):
+        self.set_http_param('AntiFog', '1')
+        self.set_http_param('AntiFogLevel', value)
+    
+    def disableAntiFog(self):
+        self.set_http_param('AntiFog', '0')
+    
+    def setScene(self, value = 'indoor'):
```
Pmichaud29: Can we use a map/enum, that way we don't have to check the source code to know valid inputs? (indoor, outdoor, etc.)

Pmichaud29 (changes requested):
Globally very good, I would add a bit of verbose for input values (scope them to an enum or any custom type, this way the user can only enter valid values and can know them without looking at the source code). Also add checks for ranges values to make sure the value we sent to the camera make senses

Pmichaud29 (approved):

PR #98, opened 2025-01-30
first implementation of aruco detection

Gabduf:


rover_auxiliary/src/aruco_detection.h
```diff
@@ -0,0 +1,94 @@
...
+#include "rover_msgs/msg/aruco.hpp"
+
+
+namespace ArucoDetectionHelpers {
+
+    constexpr uint8_t MAX_SAME_FRAME_IDS = 20;
+    constexpr cv::aruco::PREDEFINED_DICTIONARY_NAME DICT = cv::aruco::DICT_4X4_250;
+    constexpr uint8_t DELAY_BETWEEN_CAPTURE_MS = 100;
+
+    enum CameraAccessMode : uint8_t { URL = 0, ID = 1 };
```
Pmichaud29: Could be inside ImageCapture Class to remove main namespace

rover_auxiliary/src/aruco_detection.h
```diff
@@ -0,0 +1,94 @@
...
+#include <chrono>
+#include "rclcpp/rclcpp.hpp"
+#include "rover_msgs/msg/aruco.hpp"
+
+
+namespace ArucoDetectionHelpers {
+
+    constexpr uint8_t MAX_SAME_FRAME_IDS = 20;
+    constexpr cv::aruco::PREDEFINED_DICTIONARY_NAME DICT = cv::aruco::DICT_4X4_250;
+    constexpr uint8_t DELAY_BETWEEN_CAPTURE_MS = 100;
```
Pmichaud29: Could be static constexpr inside the class

rover_auxiliary/src/aruco_detection.h
```diff
@@ -0,0 +1,94 @@
...
+        uint8_t cameraID;
+        uint8_t cameraAccessMode;
+        cv::VideoCapture cap;
+
+    public:
+        ImageCapture(std::string _cameraURL, uint8_t _cameraAccessMode);
+        ImageCapture(uint8_t _cameraID, uint8_t _cameraAccessMode);
+        ~ImageCapture();
+
+        void setCameraURL(std::string);
```
Pmichaud29: Argument name is missing

rover_auxiliary/src/aruco_detection.h
```diff
@@ -0,0 +1,94 @@
...
+        std::string getCameraURL();
```
Pmichaud29: Try adding void argument to no arg func

rover_auxiliary/src/aruco_detection.h
```diff
@@ -0,0 +1,94 @@
...
+
+        bool accessStream();
+        bool manageStream();
+        cv::Mat getFrame();
+       // void getErrorFrame(cv::Mat&);  Use for debugging
```
Pmichaud29: I would combine both since getErrorFrame is interesting for example with optional argument on:

```
cv::Mat getFrame(bool showOnScreen = false); 
```

rover_auxiliary/src/aruco_detection.cpp
```diff
@@ -0,0 +1,198 @@
...
+
+using namespace ArucoDetectionHelpers;
+
+ImageCapture::ImageCapture(std::string _cameraURL, uint8_t  _cameraAccessMode) : cameraURL(_cameraURL), cameraAccessMode(_cameraAccessMode) {} 
+
+ImageCapture::ImageCapture(uint8_t  _cameraID, uint8_t  _cameraAccessMode) : cameraID(_cameraID), cameraAccessMode(_cameraAccessMode) {}
+
+ImageCapture::~ImageCapture() {
+    cap.release();
+    //cv::destroyAllWindows(); Use when debugging
```
Pmichaud29: Debugging could be a flag in constructor

rover_auxiliary/src/aruco_detection.cpp
```diff
@@ -0,0 +1,198 @@
...
+    rclcpp::shutdown();
+    return 0;
+}
+
+
+
+using namespace ArucoDetectionHelpers;
+
+ImageCapture::ImageCapture(std::string _cameraURL, uint8_t  _cameraAccessMode) : cameraURL(_cameraURL), cameraAccessMode(_cameraAccessMode) {} 
+
```
Pmichaud29: Since setCameraURL() is offered it would best practices to call it in constructor

rover_auxiliary/src/aruco_detection.cpp
```diff
@@ -0,0 +1,198 @@
...
+ImageCapture::ImageCapture(uint8_t  _cameraID, uint8_t  _cameraAccessMode) : cameraID(_cameraID), cameraAccessMode(_cameraAccessMode) {}
+
```
Pmichaud29: Argument naming should have trailing underscore ex: uint8_t cameraID_

rover_auxiliary/src/aruco_detection.cpp
```diff
@@ -0,0 +1,198 @@
...
+ImageCapture::~ImageCapture() {
+    cap.release();
+    //cv::destroyAllWindows(); Use when debugging
+}
+
+void ImageCapture::setCameraURL(std::string URL) { cameraURL = URL; }
+
+std::string ImageCapture::getCameraURL() { return cameraURL; }
+
```
Pmichaud29: Should release and reopen new stream when changing URL

rover_auxiliary/src/aruco_detection.cpp
```diff
@@ -0,0 +1,198 @@
...
+    if (!cap.isOpened()) {
+        if (!manageStream()) {    // Open camera only if it's off
+            //getErrorFrame(frame); Use when debugging
+            return frame;
+        }
+    }
+
+    cap >> frame;   // Store frame in matrix  (openCV syntax)
+    return frame;
+}
```
Pmichaud29: Return value could be a std::optional<cv::Mat> and the func could go like this, this way the "user" (class using this class) could report meaningfull feedback on error:

```
#include <optional>
std::optional<cv::Mat> ImageCapture::getFrame()
{
    if (!cap.isOpened() && !manageStream())
    {
        return std::nullopt;
    }
    else
    {
        cv::Mat frame;
        return cap >> frame;
    }
}

// Usage example:
std::optional<cv::Mat> frame = ImageCapture.getFrame()
if (!frame)
{
    RCLCPP_WARN("Error getting frame from stream");
}
```

rover_auxiliary/src/aruco_detection.cpp
```diff
@@ -0,0 +1,198 @@
...
+/* Use when debugging
+void ImageCapture::getErrorFrame(cv::Mat& frame) {
+    frame = cv::Mat::zeros(480, 640, CV_8UC3);
+    std::string error_message = "Error: Stream not found!";
+    cv::putText(frame, error_message, cv::Point(100, 240), cv::FONT_HERSHEY_SIMPLEX, 1.0, cv::Scalar(0, 0, 255), 2);
+}*/
+
+// Init frame processing when accessing camera with ID
+FrameProcessing::FrameProcessing(std::string _cameraURL, uint8_t  _cameraAccessMode) : stream(_cameraURL, _cameraAccessMode) {
+    dictionary = cv::aruco::getPredefinedDictionary(DICT);
```
Pmichaud29: I believe it should be constant

rover_auxiliary/src/aruco_detection.cpp
```diff
@@ -0,0 +1,198 @@
...
+/* Use when debugging
+cv::Mat FrameProcessing::processedFrameToShow() {
+    processFrame();
+    cv::Mat processedFrame = stream.getFrame();
+    cv::aruco::drawDetectedMarkers(processedFrame, corners, ids);
+    return processedFrame;
+}*/
+
+std::vector<uint16_t > FrameProcessing::getIds() {
+    return detectedIds;
```
Pmichaud29: Might not be necessary, processFrame also does it. Could be a cached accessor: getCachedIds()

rover_auxiliary/src/aruco_detection.cpp
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    for (const auto& [id, count] : validationCounts) {
+        if (count >= 10) {
+            validatedIds.push_back(id);
+            validationCounts[id] = 0; // Reset count after validation
+        }
+    }
+
+    std::this_thread::sleep_for(std::chrono::milliseconds(ArucoDetectionHelpers::DELAY_BETWEEN_CAPTURE_MS));
```
Pmichaud29: Ça c'est mal!

Un timer blockant dans le thread principal c'est jamais une bonne idée, surtout que tes callbacks de ros dépendent de la vitesse à laquelle "Node::spin()" est caller. (Spoiler alert dans un timer bloquant jamais).

Utiliser une timer serait mieux:
```
RoverHelpers::Timer<unsigned long, millis> timer(1000);
for (EVER)
{
    if (timer.isDone())
    {
        /* Do something at 1 Hz*/
    }
}
```

rover_auxiliary/src/aruco_detection.cpp
```diff
@@ -0,0 +1,198 @@
...
+        }
+        strike = 0;
+    }
+
+    for (const auto& [id, count] : validationCounts) {
+        if (count >= 10) {
+            validatedIds.push_back(id);
+            validationCounts[id] = 0; // Reset count after validation
+        }
+    }
```
Pmichaud29: Could use a moving average for simpler management.

rover_auxiliary/src/aruco_detection.h
```diff
@@ -0,0 +1,94 @@
...
+    public:
+        ArucoDetectionNode();
+        ~ArucoDetectionNode();
+ 
+    private:
+        void ArucoCallback();
+        rclcpp::Publisher<rover_msgs::msg::Aruco>::SharedPtr publisher_;
+        rclcpp::TimerBase::SharedPtr timer_;
+        std::unique_ptr<ArucoDetection> detection_;
+};
```
Pmichaud29: Doesn't need to be unique_ptr, could be stack allocated which is safer for heap management. (Impact would probably be minimal but still)

rover_auxiliary/src/aruco_detection.cpp
```diff
@@ -0,0 +1,198 @@
...
+
+
+ArucoDetectionNode::ArucoDetectionNode() : Node("aurco_detection_node")
+{
+    publisher_ = this->create_publisher<rover_msgs::msg::Aruco>("detected_arucos", 10);
+    timer_ = this->create_wall_timer(std::chrono::milliseconds(100), std::bind(&ArucoDetectionNode::ArucoCallback, this));
+    detection_ = std::make_unique<ArucoDetection>(0, CameraAccessMode::ID);
+}
+    
+ArucoDetectionNode::~ArucoDetectionNode() {}
```
Pmichaud29: Could be default

rover_auxiliary/src/aruco_detection.cpp
```diff
@@ -0,0 +1,198 @@
...
+        }
+    }
+
+    std::this_thread::sleep_for(std::chrono::milliseconds(ArucoDetectionHelpers::DELAY_BETWEEN_CAPTURE_MS));
+    return validatedIds;
+}
+
+
+ArucoDetectionNode::ArucoDetectionNode() : Node("aurco_detection_node")
+{
```
Pmichaud29: Type in name

rover_auxiliary/src/aruco_detection.cpp
```diff
@@ -0,0 +1,198 @@
...
+    publisher_ = this->create_publisher<rover_msgs::msg::Aruco>("detected_arucos", 10);
```
Pmichaud29: Topic name should match global architecture in rover_documentation/doc/arch...

rover_auxiliary/src/aruco_detection.cpp
```diff
@@ -0,0 +1,198 @@
...
+    timer_ = this->create_wall_timer(std::chrono::milliseconds(100), std::bind(&ArucoDetectionNode::ArucoCallback, this));
```
Pmichaud29: std::bind could be lambda function instead which is easier to read with arguments:

```
std::bind(&ArucoDetectionNode::ArucoCallback, this) -> [this](void){ this->ArucoCallback(); };
```

rover_auxiliary/src/aruco_detection.cpp
```diff
@@ -0,0 +1,198 @@
...
+    detection_ = std::make_unique<ArucoDetection>(0, CameraAccessMode::ID);
+}
+    
+ArucoDetectionNode::~ArucoDetectionNode() {}
+ 
+void ArucoDetectionNode::ArucoCallback()
```
Pmichaud29: Callback often are named: CB_"name"

rover_auxiliary/src/aruco_detection.cpp
```diff
@@ -0,0 +1,198 @@
...
+{
+    std::vector<uint16_t> detectedArucos = detection_->update();
+
+       rover_msgs::msg::Aruco msg;
+
+        msg.valid = !detectedArucos.empty();
+
```
Pmichaud29: I don't think an empty detection should be considered invalid, error reading frame could be tho

rover_auxiliary/src/aruco_detection.cpp
```diff
@@ -0,0 +1,198 @@
...
+            {
+                marker_list += std::to_string(id) + " ";
+            }
+            RCLCPP_INFO(this->get_logger(), "%s", marker_list.c_str());
+        }
+        else
+        {
+            RCLCPP_INFO(this->get_logger(), "No Aruco markers detected to publish");
+        }
+}
```
Pmichaud29: Logging is always nice, for successes cases like this use the RCLCPP_DEBUG which doesn't show by default but is logged into the log file

Pmichaud29 (changes requested):

rover_auxiliary/src/aruco_detection.h
Gabduf: done

rover_auxiliary/src/aruco_detection.h
Gabduf: done

rover_auxiliary/src/aruco_detection.h
Gabduf: done

rover_auxiliary/src/aruco_detection.h
Gabduf: done

rover_auxiliary/src/aruco_detection.h
Gabduf: done

rover_auxiliary/src/aruco_detection.cpp
Gabduf: done

rover_auxiliary/src/aruco_detection.cpp
Gabduf: done

rover_auxiliary/src/aruco_detection.cpp
Gabduf: done

rover_auxiliary/src/aruco_detection.cpp
Gabduf: done

rover_auxiliary/src/aruco_detection.cpp
Gabduf: done

rover_auxiliary/src/aruco_detection.cpp
Gabduf: the actual name of the dictionary is a constant (ex: DICT_4X4_250), but this is a necessary step to load the dict...

rover_auxiliary/src/aruco_detection.cpp
Gabduf: Now it is necessary since processFrame only returns an optional frame (used with DEBUG_MODE)

rover_auxiliary/src/aruco_detection.cpp
Gabduf: done

rover_auxiliary/src/aruco_detection.cpp
Gabduf: done

rover_auxiliary/src/aruco_detection.h
Gabduf: The detection doesn't work when changing the unique ptr to a shared ptr. I'll check why later...

rover_auxiliary/src/aruco_detection.cpp
Gabduf: done

rover_auxiliary/src/aruco_detection.cpp
Gabduf: done

rover_auxiliary/src/aruco_detection.cpp
Gabduf: no name was defined

rover_auxiliary/src/aruco_detection.cpp
Gabduf: done

rover_auxiliary/src/aruco_detection.cpp
Gabduf: done

rover_auxiliary/src/aruco_detection.cpp
Gabduf: done

rover_auxiliary/src/aruco_detection.cpp
Gabduf: done

rover_auxiliary/CMakeLists.txt
```diff
@@ -12,15 +12,23 @@ endif()
 find_package(ament_cmake REQUIRED)
 find_package(rclcpp REQUIRED)
 find_package(rover_msgs REQUIRED)
+find_package(OpenCV REQUIRED)
+find_package(std_msgs REQUIRED)
```
Pmichaud29, line 16: Necessary?

Pmichaud29: Necessary?

rover_auxiliary/src/aruco_detection/aruco_detection_node.cpp
```diff
@@ -0,0 +1,72 @@
...
+    executor.spin();
+    executor.remove_node(node);
+    rclcpp::shutdown();
+    return 0;
+}
+
+ArucoDetectionNode::ArucoDetectionNode(void): Node("aruco_detection_node")
+{
+    declare_parameter("DEBUG_MODE", false);  // Command-line argument for quick debugging
+    get_parameter("DEBUG_MODE", DEBUG_MODE);
```
Pmichaud29: Could we use agrc argv instead? If I remember correctly this seems like a ros param which should be able to be changed dynamically at run time. Using it like this gives the user the feeling it can be changed dynamically but it won't work without giving him any feedback.

rover_auxiliary/src/aruco_detection/aruco_detection_node.hpp
```diff
@@ -0,0 +1,25 @@
...
+    static constexpr uint16_t DELAY_PUBLISHER_MS = 1000;
+    static constexpr uint16_t DELAY_DETECTION_MS = 50;
+    void CB_aruco_publisher(void);
+    void CB_aruco_detection(void);
+    rclcpp::Publisher<rover_msgs::msg::Aruco>::SharedPtr _publisher;
+    rclcpp::TimerBase::SharedPtr _timerPublisher;
+    rclcpp::TimerBase::SharedPtr _timerDetection;
+    std::unique_ptr<Detection> _detection;
+    std::mutex _mutex;
+};
```
Pmichaud29: Just styling but I've added a guideline for static constexpr placement and class member/method order: (Class header styling)

```suggestion
class ArucoDetectionNode : public rclcpp::Node
{
    static constexpr uint16_t DELAY_PUBLISHER_MS = 1000;
    static constexpr uint16_t DELAY_DETECTION_MS = 50;

  public:
    ArucoDetectionNode(void);

  private:
    void CB_aruco_publisher(void);
    void CB_aruco_detection(void);
    
    bool _debugMode;
    rclcpp::Publisher<rover_msgs::msg::Aruco>::SharedPtr _publisher;
    rclcpp::TimerBase::SharedPtr _timerPublisher;
    rclcpp::TimerBase::SharedPtr _timerDetection;
    std::unique_ptr<Detection> _detection;
    std::mutex _mutex;
};
```

rover_auxiliary/src/aruco_detection/aruco_detection_node.hpp
```diff
@@ -0,0 +1,25 @@
...
+#include "detection.hpp"
+#include "rover_msgs/msg/aruco.hpp"
+
+class ArucoDetectionNode : public rclcpp::Node
+{
+  public:
+    ArucoDetectionNode(void);
+
+  private:
+    bool DEBUG_MODE;
```
Pmichaud29: Styling: _debugMode

rover_auxiliary/src/aruco_detection/aruco_detection_node.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+    _publisher = this->create_publisher<rover_msgs::msg::Aruco>("aruco_detection", 10);
+    _timerPublisher
+        = this->create_wall_timer(std::chrono::milliseconds(DELAY_PUBLISHER_MS), [this]() { this->CB_aruco_publisher(); });
+
+    _timerDetection
+        = this->create_wall_timer(std::chrono::milliseconds(DELAY_DETECTION_MS), [this]() { this->CB_aruco_detection(); });
+
+    _detection = std::make_unique<Detection>("v4l2:///dev/video0");
```
Pmichaud29: Use launch param with default value instead

rover_auxiliary/src/aruco_detection/aruco_detection_node.cpp
```diff
@@ -0,0 +1,72 @@
...
+}
+
+void ArucoDetectionNode::CB_aruco_publisher(void)
```
Pmichaud29: CB_arucoPublisher

rover_auxiliary/src/aruco_detection/aruco_detection_node.cpp
```diff
@@ -0,0 +1,72 @@
...
+{
+    std::vector<uint16_t> detectedArucos;
+
+    {
+        std::lock_guard<std::mutex> lock(_mutex);
```
Pmichaud29: Great use/implementation of multithreading, just give your mutex a name related to the protected data

rover_auxiliary/src/aruco_detection/aruco_detection_node.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+        detectedArucos = _detection->getValidatedIds();
+    }
+
+    rover_msgs::msg::Aruco msg;
+
+    for (auto id : detectedArucos)
+    {
+        msg.id.push_back(static_cast<uint8_t>(id));  // Ensure ID is cast to uint8
```
Pmichaud29: Casting from uint16_t to uint8_t without checking if the value is in the uint8_t range will lead to unexpected behavior. I would advice changing the rover_msgs::msg::Aruco to uint16_t instead. (I know the ids in the current dict can't go over 200, but the dict could be different in the future)

Also remove comments for self explainatory lines

rover_auxiliary/src/aruco_detection/aruco_detection_node.cpp
```diff
@@ -0,0 +1,72 @@
...
+}
+
+ArucoDetectionNode::ArucoDetectionNode(void): Node("aruco_detection_node")
+{
+    declare_parameter("DEBUG_MODE", false);  // Command-line argument for quick debugging
+    get_parameter("DEBUG_MODE", DEBUG_MODE);
+
+    _publisher = this->create_publisher<rover_msgs::msg::Aruco>("aruco_detection", 10);
+    _timerPublisher
+        = this->create_wall_timer(std::chrono::milliseconds(DELAY_PUBLISHER_MS), [this]() { this->CB_aruco_publisher(); });
```
Pmichaud29: Either cast the DELAY_PUBLISHER_MS into int64_t or define it with the correct type for underlying constructor (std::chrono::duration<int64_t, std::milli>)

rover_auxiliary/src/aruco_detection/image_capture.cpp
```diff
@@ -0,0 +1,79 @@
...
+{
+    if (_cap.isOpened())
+    {
+        return true;
+    }
+
+    if (!_cap.open(_cameraURL))
+    {
+        std::string message = "Could not open streaming device";
+        RCLCPP_WARN(rclcpp::get_logger("ArucoDetection"), message.c_str());
```
Pmichaud29:
```suggestion
        RCLCPP_WARN(rclcpp::get_logger("ArucoDetection"), "Could not open streaming device");
```

rover_auxiliary/src/aruco_detection/image_capture.cpp
```diff
@@ -0,0 +1,79 @@
...
+        return false;
+    }
+    return true;
+}
+
+bool ImageCapture::manageStream(std::string URL_)
```
Pmichaud29: changeStream or smtg would be clearer

rover_auxiliary/src/aruco_detection/image_capture.cpp
```diff
@@ -0,0 +1,79 @@
...
+        }
+        return true;
+    }
+    else
+    {
+        return initCam();
+    }
+}
+
+std::optional<cv::Mat> ImageCapture::getFrame(bool DEBUG_MODE)
```
Pmichaud29: use argument name convention debugMode_

Also, does this return the most recent frame?

rover_auxiliary/src/aruco_detection/image_capture.cpp
```diff
@@ -0,0 +1,79 @@
...
+    _cap >> frame;  // Store frame in matrix  (openCV syntax)
+
+    if (frame.empty())
+    {
+        return std::nullopt;
+    }
+    return frame;
+}
+
+void ImageCapture::getErrorFrame(cv::Mat& frame_)
```
Pmichaud29: When possible use a return value; in this case:
cv::Mat ImageCapture::getErrorFrame(void)

The use of ref this way is mostly when there's multiple arguments or when there's a returned error code. The use of optional is there when the function can fail.

rover_auxiliary/src/aruco_detection/frame_processing.cpp
```diff
@@ -0,0 +1,44 @@
+#include "frame_processing.hpp"
+
+// Init frame processing when accessing camera with ID
+FrameProcessing::FrameProcessing(std::string cameraURL_): _stream(cameraURL_)
+{
+    _dictionary = cv::aruco::getPredefinedDictionary(DICT);
+    _detectorParams = cv::aruco::DetectorParameters::create();
```
Pmichaud29: Not used

rover_auxiliary/src/aruco_detection/frame_processing.cpp
```diff
@@ -0,0 +1,44 @@
+#include "frame_processing.hpp"
+
+// Init frame processing when accessing camera with ID
+FrameProcessing::FrameProcessing(std::string cameraURL_): _stream(cameraURL_)
+{
+    _dictionary = cv::aruco::getPredefinedDictionary(DICT);
```
Pmichaud29: Could be const and inited like this:
```cpp
FrameProcessing::FrameProcessing(std::string cameraURL_): _stream(cameraURL_), _dictionary(cv::aruco::getPredefinedDictionary(DICT))
{}
```

rover_auxiliary/src/aruco_detection/frame_processing.cpp
```diff
@@ -0,0 +1,44 @@
...
+    _detectorParams = cv::aruco::DetectorParameters::create();
+}
+
+FrameProcessing::~FrameProcessing(void) {}
```
Pmichaud29: Don't declare/define or declare default instead

rover_auxiliary/src/aruco_detection/frame_processing.cpp
```diff
@@ -0,0 +1,44 @@
...
+    {
+        _detectedIds.push_back(_ids.at<uint16_t>(i, 0));
+    }
+
+    if (DEBUG_MODE)
+    {
+        cv::aruco::drawDetectedMarkers(frame.value(), _corners, _ids);
+        return frame;
+    }
+    empty = _detectedIds.empty();
```
Pmichaud29: Wouldn't we want to do this directly in the FrameProcessing::IdsEmpty(void) func?

ex:
```
bool FrameProcessing::IdsEmpty(void)
{
    return _detectedIds.empty();
}
```

rover_auxiliary/src/aruco_detection/detection.hpp
```diff
@@ -0,0 +1,26 @@
...
+  public:
+    Detection(std::string cameraURL_);
+    std::vector<uint16_t> detect(bool DEBUG_MODE);
+    void update(bool DEBUG_MODE);
+    std::vector<uint16_t> getValidatedIds(void);
+
+  private:
+    static constexpr float VALIDATION_THRESHOLD = 0.75;
+    static constexpr uint64_t DELAY_BETWEEN_CAPTURE_MS = 100;
+    FrameProcessing _processedFrame;
```
Pmichaud29: naming is a bit weird, frameProcessing or camera or smtg else would make a bit sense

rover_auxiliary/src/aruco_detection/frame_processing.hpp
```diff
@@ -0,0 +1,30 @@
...
+
+  private:
+    static constexpr uint8_t MAX_SAME_FRAME_IDS = 20;
+    static constexpr cv::aruco::PREDEFINED_DICTIONARY_NAME DICT = cv::aruco::DICT_4X4_250;
+    ImageCapture _stream;
+    std::vector<uint16_t> _detectedIds;
+    std::vector<std::vector<cv::Point2f>> _corners;
+    cv::Mat _ids;
+    cv::Ptr<cv::aruco::Dictionary> _dictionary;
+    cv::Ptr<cv::aruco::DetectorParameters> _detectorParams;
```
Pmichaud29: Not used

rover_auxiliary/src/aruco_detection/frame_processing.hpp
```diff
@@ -0,0 +1,30 @@
+#ifndef __FRAME_PROCESSING_HPP__
+#define __FRAME_PROCESSING_HPP__
+
+#include "image_capture.hpp"
+#include <optional>
+#include <vector>
+
+class FrameProcessing
```
Pmichaud29: This class is a bit confusing for me, it's defined like a vector wrapper around the detected ids, but the naming logic tend to mean it's a processing class. This hides the fact that you have the periodically call processFrame to get valid detectedIds values.

Basically the flow for the user is a bit confusing:
I personnally would keep the processFrame method but it would return directly an optional vector, the user could then do:

```
auto ids = frameProcessing.processFrame();
if (!ids)
{
    // Error processing frame
}
else if (ids.get().empty())
{
    // No ID detected
}
else
{
    // Found id list
}
``

**Review by Pmichaud29 — CHANGES_REQUESTED** [review]
Very good version, there are still some small elements to work on but nothing major.

File: rover_auxiliary/CMakeLists.txt (hunk 23 above)
Comment (Gabduf) [reply rover_auxiliary/CMakeLists.txt]: not std_msgs, done

File: rover_auxiliary/src/aruco_detection/aruco_detection_node.cpp (hunk 24 above)
Comment (Gabduf) [reply rover_auxiliary/src/aruco_detection/aruco_detection_node.cpp]: Changed it for a launch param with default to false so we can use with launch

File: rover_auxiliary/src/aruco_detection/aruco_detection_node.hpp (hunk 25 above)
Comment (Gabduf) [reply rover_auxiliary/src/aruco_detection/aruco_detection_node.hpp]: done

File: rover_auxiliary/src/aruco_detection/aruco_detection_node.hpp (hunk 26 above)
Comment (Gabduf) [reply rover_auxiliary/src/aruco_detection/aruco_detection_node.hpp]: done

File: rover_auxiliary/src/aruco_detection/aruco_detection_node.cpp (hunk 27 above)
Comment (Gabduf) [reply rover_auxiliary/src/aruco_detection/aruco_detection_node.cpp]: done

File: rover_auxiliary/src/aruco_detection/aruco_detection_node.cpp (hunk 28 above)
Comment (Gabduf) [reply rover_auxiliary/src/aruco_detection/aruco_detection_node.cpp]: done

File: rover_auxiliary/src/aruco_detection/aruco_detection_node.cpp (hunk 29 above)
Comment (Gabduf) [reply rover_auxiliary/src/aruco_detection/aruco_detection_node.cpp]: done

File: rover_auxiliary/src/aruco_detection/aruco_detection_node.cpp (hunk 30 above)
Comment (Gabduf) [reply rover_auxiliary/src/aruco_detection/aruco_detection_node.cpp]: done

File: rover_auxiliary/src/aruco_detection/aruco_detection_node.cpp (hunk 31 above)
Comment (Gabduf) [reply rover_auxiliary/src/aruco_detection/aruco_detection_node.cpp]: done

File: rover_auxiliary/src/aruco_detection/image_capture.cpp (hunk 32 above)
Comment (Gabduf) [reply rover_auxiliary/src/aruco_detection/image_capture.cpp]: done

File: rover_auxiliary/src/aruco_detection/image_capture.cpp (hunk 33 above)
Comment (Gabduf) [reply rover_auxiliary/src/aruco_detection/image_capture.cpp]: done

File: rover_auxiliary/src/aruco_detection/image_capture.cpp (hunk 34 above)
Comment (Gabduf) [reply rover_auxiliary/src/aruco_detection/image_capture.cpp]: Done,

Yes it does. In  "_cap >> frame" , _cap is a cv::VideoCapture instance that automatically reads a new frame before storing it into frame (when using the >> operator)... Updated the comment for clarity.

File: rover_auxiliary/src/aruco_detection/image_capture.cpp (hunk 35 above)
Comment (Gabduf) [reply rover_auxiliary/src/aruco_detection/image_capture.cpp]: Done

File: rover_auxiliary/src/aruco_detection/frame_processing.cpp (hunk 36 above)
Comment (Gabduf) [reply rover_auxiliary/src/aruco_detection/frame_processing.cpp]: Right. Now the params variable is passed to the detect function. Gives us more flexibility in the future to tweak the detection.

File: rover_auxiliary/src/aruco_detection/frame_processing.cpp (hunk 37 above)
Comment (Gabduf) [reply rover_auxiliary/src/aruco_detection/frame_processing.cpp]: done

File: rover_auxiliary/src/aruco_detection/frame_processing.cpp (hunk 38 above)
Comment (Gabduf) [reply rover_auxiliary/src/aruco_detection/frame_processing.cpp]: done

File: rover_auxiliary/src/aruco_detection/frame_processing.cpp (hunk 39 above)
Comment (Gabduf) [reply rover_auxiliary/src/aruco_detection/frame_processing.cpp]: done

File: rover_auxiliary/src/aruco_detection/detection.hpp (hunk 40 above)
Comment (Gabduf) [reply rover_auxiliary/src/aruco_detection/detection.hpp]: done

File: rover_auxiliary/src/aruco_detection/frame_processing.hpp (hunk 41 above)
Comment (Gabduf) [reply rover_auxiliary/src/aruco_detection/frame_processing.hpp]: now it is

File: rover_auxiliary/src/aruco_detection/frame_processing.hpp (hunk 42 above)
Comment (Gabduf) [reply rover_auxiliary/src/aruco_detection/frame_processing.hpp]: The function has been renamed "updateDetection". This makes it clearer for the user that it needs to be called periodically to work. The class has been renamed "processFrame" to make it clear that all it does is process one frame at a time. 

So the flow goes like this:
ImageCapture (capture an image) -> ProcessFrame (process the frame) -> Detection (handles the detection with its update functions, that itself calls the ProcessFrame::updateDetection...) -> arucoDetectionNode (handles the ROS integration)

File: rover_auxiliary/launch/auxiliary_rover.launch.py [hunk 43]
```diff
@@ -21,8 +21,17 @@ def generate_launch_description():
...
+        executable="aruco_detection",
+        name="aruco_detection",
+        parameters=[{'default_cam':'rtsp://192.168.1.18:554/1/h264major' },
+                    {'debug_mode': False}]
+    )
+    
     ld.add_action(node_compass_calibrator)
     ld.add_action(node_light_control)
-    # ld.add_action(node_aruco)
+    ld.add_action(node_aruco)
```
Pmichaud29: On veut le mettre à on par default? Ou c'est avec le gui qu'on controle?

rover_auxiliary/src/aruco_detection/aruco_detection_node.cpp
```diff
@@ -0,0 +1,88 @@
...
+    this->getParams(argc, argv);
+
+    _publisher = this->create_publisher<rover_msgs::msg::Aruco>("/rover/video/aruco", 10);
+    _timerPublisher
+        = this->create_wall_timer(std::chrono::milliseconds(DELAY_PUBLISHER_MS), [this](void) { this->CB_arucoPublisher(); });
+
+    _timerDetection
+        = this->create_wall_timer(std::chrono::milliseconds(DELAY_DETECTION_MS), [this](void) { this->CB_arucoDetection(); });
+
+    _detection = std::make_unique<Detection>(_camURL);
```
Pmichaud29, line 26: Si tu utilise un pointer (smart ou pas), faut qu'il soit initialisé avant que les callbacks soient callé ou que dans les callbacks tu valide qu'il existe. Un smart pointer peut quand même être nullptr. Si c'est pas nécéssaire de faire un unique_ptr / shared_ptr tes mieux de garder ça sur le stack.

rover_auxiliary/src/aruco_detection/aruco_detection_node.cpp
```diff
@@ -0,0 +1,88 @@
...
+}
+
+void ArucoDetectionNode::getParams(int argc, char** argv)
+{
+    this->declare_parameter<bool>("debug_mode", false);
+    this->get_parameter("debug_mode", _debugMode);
+
+    this->declare_parameter<std::string>("default_cam", "rtsp://192.168.1.18:554/1/h264major");
```
Pmichaud29: 192.168.144.30

rover_auxiliary/src/aruco_detection/aruco_detection_node.cpp
```diff
@@ -0,0 +1,88 @@
...
+    this->get_parameter("default_cam", _camURL);
+
+    if (argc > 1)
+    {
+        if (argv[1][0] == 'd')  // quick debugging with CLI arg 'd' or 'debug'
```
Pmichaud29: Useless comments. Ça pourrait être parser comme ça pour chaque params pour éviter le "declare_parameter" qui crée un param ros. Pour cette version c'est chill, mais à garder en tête

rover_auxiliary/src/aruco_detection/aruco_detection_node.cpp
```diff
@@ -0,0 +1,88 @@
...
+}
+
+void ArucoDetectionNode::CB_arucoPublisher(void)
+{
+    std::vector<uint16_t> detectedArucos;
+
+    {
+        std::lock_guard<std::mutex> lock(_detectedArucosMutex);
+
+        detectedArucos = _detection->getValidatedIds();
```
Pmichaud29, line 54: if (!_detection)
{
  ...
}

rover_auxiliary/src/aruco_detection/aruco_detection_node.cpp
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    rover_msgs::msg::Aruco msg;
+
+    for (auto id : detectedArucos)
```
Pmichaud29: const auto& id est une meilleur pratique quand c'est possible (évite la copie et s'assure de pas changer detectedArucos nn plus)

rover_auxiliary/src/aruco_detection/aruco_detection_node.cpp
```diff
@@ -0,0 +1,88 @@
...
+    if (_debugMode)
+    {
+        if (!detectedArucos.empty())
+        {
+            std::string marker_list = "Publishing detected Aruco markers: ";
+            for (auto id : detectedArucos)
+            {
+                marker_list += std::to_string(id) + " ";
+            }
+            RCLCPP_INFO(this->get_logger(), "%s", marker_list.c_str());
```
Pmichaud29, line 80: Pourrait être intéressant en RCLCPP_DEBUG tous le temps et nn juste en debug. Peut être change le logging level avec debug à on

rover_auxiliary/src/aruco_detection/aruco_detection_node.hpp
```diff
@@ -0,0 +1,30 @@
+#ifndef __ARUCO_DETECTION_NODE_HPP__
+#define __ARUCO_DETECTION_NODE_HPP__
+
+#include "detection.hpp"
+#include "rover_msgs/msg/aruco.hpp"
+
+class ArucoDetectionNode : public rclcpp::Node
+{
+    static constexpr uint64_t DELAY_PUBLISHER_MS = 1000;
+    static constexpr uint64_t DELAY_DETECTION_MS = 50;
```
Pmichaud29: Bonne pratique:
1'000UL
50UL

rover_auxiliary/src/aruco_detection/image_capture.cpp
```diff
@@ -0,0 +1,78 @@
...
+    initCam();
+}
+
+ImageCapture::~ImageCapture(void)
+{
+    _cap.release();
+    cv::destroyAllWindows();
+}
+
+bool ImageCapture::initCam()
```
Pmichaud29: bonne pratique:
bool ImageCapture::initCam(void)

rover_auxiliary/src/aruco_detection/image_capture.cpp
```diff
@@ -0,0 +1,78 @@
...
+    _cap >> frame;  // Updates and stores new frame (openCV syntax)
+
+    if (frame.empty())
+    {
+        return std::nullopt;
+    }
+    return frame;
+}
+
+cv::Mat ImageCapture::getErrorFrame()
```
Pmichaud29: cv::Mat ImageCapture::getErrorFrame(void)

rover_auxiliary/src/aruco_detection/image_capture.hpp
```diff
@@ -0,0 +1,25 @@
...
+class ImageCapture
+{
+  public:
+    ImageCapture(std::string cameraURL_);
+    ~ImageCapture(void);
+
+    bool changeStream(std::string URL_);
+    std::optional<cv::Mat> getFrame(bool debugMode_);
+    cv::Mat getErrorFrame();
+    bool initCam();
```
Pmichaud29: void arg

rover_auxiliary/src/aruco_detection/process_frame.cpp
```diff
@@ -0,0 +1,39 @@
...
+{
+    _detectorParams = cv::aruco::DetectorParameters::create();
+}
+
+std::optional<cv::Mat> ProcessFrame::updateDetection(bool debugMode_)
+{
+    std::optional<cv::Mat> frame = _stream.getFrame(debugMode_);
+    if (!frame)
+    {
+        RCLCPP_WARN(rclcpp::get_logger("ArucoDetection"), "Error getting frame from stream");
```
Pmichaud29, line 13: Doit retourner sinon ça va casser à la ligne 15:
cv::aruco::detectMarkers(frame.value(), DICTIONNARY, _corners, _ids, _detectorParams);

Frame me semble un ptr

rover_auxiliary/src/aruco_detection/process_frame.hpp
```diff
@@ -0,0 +1,33 @@
...
+#include "image_capture.hpp"
+#include <optional>
+#include <vector>
+
+class ProcessFrame
+{
+    static constexpr uint8_t MAX_SAME_FRAME_IDS = 20;
+    static constexpr cv::aruco::PREDEFINED_DICTIONARY_NAME DICT = cv::aruco::DICT_4X4_250;
+
+    const cv::Ptr<cv::aruco::Dictionary> DICTIONNARY;
```
Pmichaud29: Devrait être les autres membres privés

Pmichaud29 (changes requested):
Commentaires très mineurs sauf pour les ptrs, après ça devrait bon pour merge

rover_auxiliary/launch/auxiliary_rover.launch.py
Gabduf: Je suppose GUI

rover_auxiliary/src/aruco_detection/aruco_detection_node.cpp
Gabduf: done

rover_auxiliary/src/aruco_detection/aruco_detection_node.cpp
Gabduf: Jpense les pointeurs vont peut-être simplifier pour plus tard avec les features de ajouter/kill des detections sur differentes cams? 

Pis ok done

rover_auxiliary/src/aruco_detection/aruco_detection_node.cpp
Gabduf: done

rover_auxiliary/src/aruco_detection/aruco_detection_node.cpp
Gabduf: done

rover_auxiliary/src/aruco_detection/aruco_detection_node.cpp
Gabduf: done

rover_auxiliary/src/aruco_detection/aruco_detection_node.cpp
Gabduf: done

rover_auxiliary/src/aruco_detection/aruco_detection_node.hpp
Gabduf: done

rover_auxiliary/src/aruco_detection/image_capture.cpp
Gabduf: done

rover_auxiliary/src/aruco_detection/image_capture.cpp
Gabduf: done

rover_auxiliary/src/aruco_detection/image_capture.hpp
Gabduf: done

rover_auxiliary/src/aruco_detection/process_frame.cpp
Gabduf: done

rover_auxiliary/src/aruco_detection/process_frame.hpp
Gabduf: done

rover_auxiliary/src/aruco_detection/aruco_detection_node.cpp
Pmichaud29: Très fair

Pmichaud29 (approved):

PR #101, opened 2025-03-02
Gstreamer RTSP widget for UI

ChristopherLajoie:


rover_gui/src/QRtspPlayer/QGStreamerWorker.cpp
```diff
@@ -0,0 +1,190 @@
+#include "QGStreamerWorker.hpp"
+#include "QLoggingMacros.hpp"
+#include <QDebug>
+#include <gst/app/gstappsink.h>
+#include <gst/gst.h>
+#include <gst/video/videooverlay.h>
+
+static void on_gst_error_message(GstBus* bus, GstMessage* msg, gpointer user_data)
```
Pmichaud29, line 12: Comment unused args instead of QUNUSED + func name camel case ex:

```cpp
static void onGSTErrorMsg(GstBus* /*bus*/, GstMessage* msg_, gpointer userData_)
```

rover_gui/src/QRtspPlayer/QGStreamerWorker.cpp
```diff
@@ -0,0 +1,190 @@
...
+{
+    Q_UNUSED(bus);
+    auto* worker = static_cast<GStreamerWorker*>(user_data);
```
Pmichaud29, line 15: Validate if user_data is valid before casting

reinterpret cast?

rover_gui/src/QRtspPlayer/QGStreamerWorker.cpp
```diff
@@ -0,0 +1,190 @@
...
+
+    GError* err = nullptr;
+    gchar* debug = nullptr;
+    gst_message_parse_error(msg, &err, &debug);
+
+    QString errorMsg = QString("%1").arg(err ? err->message : "Unknown Error");
```
Pmichaud29, line 21: Validate if ptr are set

rover_gui/src/QRtspPlayer/QGStreamerWorker.cpp
```diff
@@ -0,0 +1,190 @@
...
+
+    LOG_DEBUG("GStreamer", errorMsg);
```
Pmichaud29, line 23: Use Ros logging, can be done through this LOG interface

Pmichaud29 (changes requested):
What we talked in call

PR #111, opened 2025-04-11
Arm control

emilesavoie:


rover_arm/include/cartesian_controller.hpp
```diff
@@ -0,0 +1,206 @@
+#ifndef __CARTESIAN_CONTROLLER_HPP__
+#define __CARTESIAN_CONTROLLER_HPP__
+
+#include "robot_controller.hpp"
+#include "Eigen/Dense"
+
+static constexpr uint8_t CARTESIAN_JOINTS = 3;
```
Pmichaud29: maybe use eLAST in enum class eCartesian instead, this way it's a bit more modular.
Please but local constant inside scope (in this case, i believe inside CartesianController)

rover_arm/include/robot_controller.hpp
```diff
@@ -0,0 +1,60 @@
...
+    virtual ~RobotController() = default;
+    RobotController(std::initializer_list<uint8_t> joints_)
+    {
+        _nJoints = joints_.size();
+        _joints.assign(joints_.begin(), joints_.end());
+    }
+
+    virtual std::array<float, ALL_JOINTS> setCmd(std::array<float, ALL_INPUTS> inputArray_) = 0;
+
+    static float getMaxVelocity(uint8_t joint)
```
Pmichaud29: a reason why you don't take a eJointIndex instead as parameter? This way this could be constexpr and would be clearer to the user

rover_arm/include/cartesian_controller.hpp
```diff
@@ -0,0 +1,206 @@
...
+        std::array<float, CARTESIAN_JOINTS> scaledVelocities = this->scaleVelocities(velocityArray);
+
+        jointCommands[TO_UNDERLYING(eJointIndex::JL)] = scaledVelocities[TO_UNDERLYING(eJointIndex::JL)];
+        jointCommands[TO_UNDERLYING(eJointIndex::J1)] = scaledVelocities[TO_UNDERLYING(eJointIndex::J1)];
+        jointCommands[TO_UNDERLYING(eJointIndex::J2)] = scaledVelocities[TO_UNDERLYING(eJointIndex::J2)];
+
+        return jointCommands;
+    }
+
+    std::array<float, CARTESIAN_JOINTS * CARTESIAN_JOINTS> computeJacobian(std::array<float, ALL_JOINTS> currentJointPosition_)
```
Pmichaud29: Could maybe be enums:

```cpp
enum class CARTESIAN_R
{
    X = 0,
    Y,
    Z,
    eLAST,
}

enum class CARTESIAN_Q
{
    Q0 = 0,
    Q1,
    Q2,
    eLAST,
}

// Then: 
std::array<float, TO_UNDERLYING(CARTESIAN_R::eLAST) * TO_UNDERLYING(CARTESIAN_Q::eLAST)> computeJacobian...

// Which would lead clearer _jacobian index:
_jacobian[CARTESIAN_R::X * CARTESIAN_JOINTS::Q0]
_jacobian[CARTESIAN_R::X * CARTESIAN_JOINTS::Q1]
_jacobian[CARTESIAN_R::X * CARTESIAN_JOINTS::Q2]

_jacobian[CARTESIAN_R::Y * CARTESIAN_JOINTS::Q0]
...
```

This would also remove the constant "magic values" at top of the file ()

rover_arm/include/robot_controller.hpp
```diff
@@ -0,0 +1,60 @@
...
+#include "rovus_lib/timer.hpp"
+#include "rovus_lib/macros.h"
+
+#include <stdint.h>
+#include <vector>
+#include <map>
+#include <initializer_list>
+
+static constexpr uint8_t ALL_JOINTS = 6;
+static constexpr uint8_t ALL_INPUTS = 20;
```
Pmichaud29: Define enum class for both and use the "eLAST" element to retrieve the size

rover_arm/include/keybinding.hpp
```diff
@@ -1,37 +1,61 @@
...
 #define __KEYBINDING__
 
-#include "rover_msgs/msg/arm_msg.hpp"
+#include "arm_configuration.hpp"
 
-// Driver
-#define PHIL
-
-namespace KEYBINDING
+namespace KEYBINDINGS_EMILE
```
Pmichaud29: KEYBINDINGS::EMILE ?

rover_arm/include/keybinding.hpp
```diff
@@ -1,37 +1,61 @@
 #ifndef __KEYBINDING__
 #define __KEYBINDING__
```
Pmichaud29: KEYBINDING_HPP

rover_msgs/msg/ArmMsg.msg
```diff
@@ -1,9 +1,8 @@
...
-uint8 GRIPPER_ROT= 5
-uint8 GRIPPER_CLOSE = 6
+uint8 J1 = 1
+uint8 J2 = 2
+uint8 GRIPPER_TILT = 3
+uint8 GRIPPER_ROT= 4
+uint8 GRIPPER_CLOSE = 5
 
-float32[7] data
+float32[6] data
```
Pmichaud29: Keep J0, with the new way you handle controller, they now can individually exclude it. This way it's going to be a lot easier when we include the rotation

rover_can/src/can_master.cpp
Pmichaud29, line 1: Keep the same msgs definition as before. There shouldn't be modifications to this file

rover_arm/include/joy_controller.hpp
```diff
@@ -0,0 +1,60 @@
...
+    std::map<uint8_t, bool> _buttonStates = {{rover_msgs::msg::Joy::CROSS_UP, false},
+                                             {rover_msgs::msg::Joy::CROSS_DOWN, false},
+                                             {rover_msgs::msg::Joy::CROSS_RIGHT, false},
+                                             {rover_msgs::msg::Joy::CROSS_LEFT, false},
+                                             {rover_msgs::msg::Joy::A, false},
+                                             {rover_msgs::msg::Joy::B, false},
+                                             {rover_msgs::msg::Joy::X, false},
+                                             {rover_msgs::msg::Joy::Y, false},
+                                             {rover_msgs::msg::Joy::L1, false},
+                                             {rover_msgs::msg::Joy::L2, false}};
```
Pmichaud29: Define an enum class type here which maps the uint8_t value from rover_msgs::msg::Joy to be used everywhere else in your code:

```cpp
enum class eJoyMap /* or any better name */ : uint8_t
{
    CROSS_UP = rover_msgs::msg::Joy::CROSS_UP,
    CROSS_DOWN = rover_msgs::msg::Joy::CROSS_DOWN,
    CROSS_RIGHT = rover_msgs::msg::Joy::CROSS_LEFT,
    A = rover_msgs::msg::Joy::A,
    ...
    eLAST,
};

```

Pmichaud29 (changes requested):
C'est globalement super bien pis la nouvelle organisation fait vrm beaucoup de sens et elle est bien fait. Je dirais qu'il manque juste un peu d'amour au niveau du mapping des bindings et des directions. À certaines place ça utilise des enums, mais après ceux-ci sont transformé en uint8_t dans certain object, mais pas d'autre.

En gros juste revoir les places ou ça utilises des uint8_t comme index et les modifier pour utiliser des enums.  (Ça inclue, les array, les arguments de fonction et les constantes en top de fichier).

Aussi le J0 devrait pas être enlevé des définitions globales, seulement des controlleurs qui ne l'utilise pas. Par example le controlleur joint devrait avoir des bindings pour le J0, même si physiquement il ne bouge pas.

rover_arm/include/robot_controller.hpp
emilesavoie: Done

rover_arm/include/robot_controller.hpp
emilesavoie: Done

rover_arm/include/keybinding.hpp
emilesavoie: Done

rover_arm/include/keybinding.hpp
emilesavoie: Done

rover_msgs/msg/ArmMsg.msg
emilesavoie: Done

rover_arm/include/joy_controller.hpp
emilesavoie: Done

rover_arm/include/cartesian_controller.hpp
emilesavoie: Done

rover_arm/include/arm_configuration.hpp
```diff
@@ -1,6 +1,18 @@
 #ifndef _ARM_CONFIGURATION__
 #define _ARM_CONFIGURATION__
```
Pmichaud29, line 2: ARM_CONFIGURATION_HPP

rover_arm/include/arm_configuration.hpp
```diff
@@ -39,56 +53,69 @@ namespace ARM_CONFIGURATION
...
+    }      // namespace GRIPPER_CLOSE
+
     // GRIPPER_CLOSE not defined because no speed control available in current arm configuration
 }  // namespace ARM_CONFIGURATION
 
 // Length for MotionGenesis kinematics, defined for quicker back and forth,
 // see doc or motion genesis code for bases reference
-constexpr float J0x = ARM_CONFIGURATION::J0::LEN::x;
-constexpr float J0y = ARM_CONFIGURATION::J0::LEN::y;
-constexpr float J0z = ARM_CONFIGURATION::J0::LEN::z;
```
Pmichaud29, line 91: Might want to keep those even if ununsed

rover_arm/include/robot_controller.hpp
```diff
@@ -0,0 +1,62 @@
...
+        _joints(joints_),
+        _joyController()
+    {
+    }
+
+    virtual std::array<float, TO_UNDERLYING(eJointIndex::eLAST)> setCmd(
+        std::array<float, TO_UNDERLYING(eJoyInput::eLAST)> inputArray_)
+        = 0;
+
+    static float getMaxVelocity(eJointIndex joint_)
```
Pmichaud29, line 30: set constexpr since all value are constexpr.

rover_arm/include/robot_controller.hpp
```diff
@@ -0,0 +1,62 @@
...
+            case eJointIndex::J1:
+                return ARM_CONFIGURATION::J1::MAX_VELOCITY;
+            case eJointIndex::J2:
+                return ARM_CONFIGURATION::J2::MAX_VELOCITY;
+            case eJointIndex::GRIPPER_TILT:
+                return ARM_CONFIGURATION::GRIPPER_TILT::MAX_VELOCITY;
+            case eJointIndex::GRIPPER_ROT:
+                return ARM_CONFIGURATION::GRIPPER_ROT::MAX_VELOCITY;
+            default:
+                return 0.0F;
```
Pmichaud29, line 47: Don't define default case in closed enum to enable compilation warning on unhandled cases. If the "user" provides a explicitely casted custom value it's his problem

rover_arm/include/robot_controller.hpp
```diff
@@ -0,0 +1,62 @@
...
+        }
+    }
+
+    uint8_t getNJoints(void)
+    {
+        return _nJoints;
+    }
```
Pmichaud29, line 54: Use underlying type of eJointIndex instead of uin8_t. This also mean you should specify which type eJointIndex is like this:

enum class eJointIndex: size_t
{
    ...
}; 

On linux you've got plenty of memory so uint8_t is fine but size_t is generally recommended.

rover_arm/include/robot_controller.hpp
```diff
@@ -0,0 +1,62 @@
...
+
+  protected:
+    uint8_t _nJoints;
+    std::vector<eJointIndex> _joints;
```
Pmichaud29, line 58: Reason why _nJoints exist and we don't use _joints.size() instead?

Also I thing _joints isn't used

rover_arm/include/robot_controller.hpp
```diff
@@ -0,0 +1,62 @@
...
+    JoyController _joyController;
```
Pmichaud29, line 59: Necessary here? Shouldn't it be the RobotController's child's choice to include one instead?

rover_arm/include/keybinding.hpp
```diff
@@ -1,37 +1,86 @@
...
+    EXT2 = rover_msgs::msg::Joy::EXT2,
+    eLAST
+};
 
-namespace KEYBINDING
+namespace KEYBINDINGS
 {
-#if defined(PHIL)
-    constexpr uint8_t DEADMAN_SWITCH = rover_msgs::msg::Joy::L1;
+    namespace EMILE
```
Pmichaud29, line 34: With the current implementation it won't be possible to switch pilots bindings. It's fine for this PR but don't include any pilots in this case.

rover_arm/include/joy_controller.hpp
```diff
@@ -0,0 +1,60 @@
...
+#include <stdint.h>
+#include <vector>
+#include <map>
+#include <initializer_list>
+
+class JoyController
+{
+  public:
+    ~JoyController() = default;
+    JoyController() = default;
```
Pmichaud29, line 18: Not needed

rover_arm/include/joy_controller.hpp
```diff
@@ -0,0 +1,60 @@
...
+            _buttonStates[buttonId] = false;
+            return false;
+        }
+        return false;
+    }
+
+    uint8_t getNJoints(void)
+    {
+        return _nJoints;
+    }
```
Pmichaud29, line 43: Seems out of place, it's not used

rover_arm/include/joy_controller.hpp
```diff
@@ -0,0 +1,60 @@
...
+
+  protected:
+    uint8_t _nJoints;
+    std::vector<uint8_t> _joints;
```
Pmichaud29, line 47: Seems out of place, it's not used

rover_arm/include/joy_controller.hpp
```diff
@@ -0,0 +1,60 @@
...
+    std::map<eJoyInput, bool> _buttonStates = {{eJoyInput::CROSS_UP, false},
```
Pmichaud29, line 48: Shouldn't this include all possible buttons of eJoyInput? It feels limiting to not be able to map buttons to other EXT0,1,2 for example. Also isSelected is hiding to the user which eJoyInput is not mapped/tracked. If there's a specific reason why it's limited to only these inputs, consider declaring a new enum with only supported values to let the user know.

rover_arm/include/joy_controller.hpp
```diff
@@ -0,0 +1,60 @@
...
+            _buttonStates[buttonId] = true;
+            return true;
+        }
+        else if (!isPressed(buttonValue_) && _buttonStates[buttonId])
+        {
+            _buttonStates[buttonId] = false;
+            return false;
+        }
+        return false;
+    }
```
Pmichaud29, line 38: Not clear from name what isSelected is meant to be doing. It feels the same isPressed from external point of view

rover_arm/include/joy_controller.hpp
```diff
@@ -0,0 +1,60 @@
...
+
+class JoyController
+{
+  public:
+    ~JoyController() = default;
+    JoyController() = default;
+
+    bool isPressed(float buttonValue_)
+    {
+        return !IN_ERROR(buttonValue_, 0.01F, 0.0F);
```
Pmichaud29, line 22: Magic values should constexpr at top of class

rover_arm/include/joy_controller.hpp
Pmichaud29, line 1: Maybe reconsider this class and it's implications, it seems to be keeping a legacy feature from the old version of the code but feels out of place. The isPressed doesn't need a class scope (namespace would do) and the isSelected() feels out of place. If you want we can call to discuss it together to find a way to do this

rover_arm/include/joint_controller.hpp
```diff
@@ -0,0 +1,109 @@
...
+    {
+        if (joints_.size() != 0)
+        {
+            _currentControlledJoint = *joints_.begin();
+        }
+        else
+        {
+            _currentControlledJoint = eJointIndex::eLAST;
+        }
+    }
```
Pmichaud29, line 21:
```suggestion
    JointController(std::initializer_list<eJointIndex> joints_):
        RobotController(joints_),
        _currentControlledJoint(eJointIndex::eLAST)
    {
        if (joints_.size() != 0)
        {
            _currentControlledJoint = *joints_.begin();
        }
    }
```

rover_arm/include/joint_controller.hpp
```diff
@@ -0,0 +1,109 @@
...
+
+    std::array<float, TO_UNDERLYING(eJointIndex::eLAST)> setCmd(
+        std::array<float, TO_UNDERLYING(eJoyInput::eLAST)> inputArray_) override
```
Pmichaud29, line 24: review method naming, this is more a getJointCmdFromInput type of method than a setCmd.

rover_arm/include/gripper_controller.hpp
```diff
@@ -0,0 +1,66 @@
+#ifndef __GRIPPER_CONTROLLER_HPP__
+#define __GRIPPER_CONTROLLER_HPP__
+
+#include "robot_controller.hpp"
+
+class GripperController : public RobotController
```
Pmichaud29, line 6: We'll maybe only want to control the actual gripping inside of this class and not the Rot/Tilt

Pmichaud29 (changes requested):

PR #113, opened 2025-04-18
Service for ddb control

AraaRobot:
Two services allowing to change the state, mode or PWM values for DDB banks (Different features per bank)

Bank 0:
-Changing state of the desired channel to ON, OFF or PWM.
-Change current values of duty cycle and frequency.

Bank 1:
-Change state of the desired channel to ON or OFF.


rover_auxiliary/src/DDB/ddb_control_node.hpp
```diff
@@ -0,0 +1,57 @@
...
+#include "rovus_lib/macros.h"
+
+#include "rclcpp/rclcpp.hpp"
+#include "rover_msgs/srv/ddb_control.hpp"
+
+class DDBControlNode : public rclcpp::Node
+{
+    static constexpr size_t MAX_CHANNELS = 4;
+
+    enum class eToggleState : size_t
```
Pmichaud29, line 14: I would call it eOutputState

Pmichaud29: I would use "Output" terminology instead of "Toggle"

rover_auxiliary/src/DDB/ddb_control_node.hpp
```diff
@@ -0,0 +1,57 @@
...
+        FIX,
+        PWM,
+    };
+
+    struct sChannelInfo
+    {
+        eToggleState state = eToggleState::ON;
+        eToggleMode mode = eToggleMode::FIX;
+        uint8_t dutyCycle = 0;
+        float frequency = 1.0;
```
Pmichaud29: Set a freq of 0Hz by default for eToggleMode::FIX channels, 0Hz would confirm constant.

rover_auxiliary/src/DDB/ddb_control_node.hpp
```diff
@@ -0,0 +1,57 @@
...
+
+#include "rovus_lib/camera_info.hpp"
+#include "rovus_lib/macros.h"
+
+#include "rclcpp/rclcpp.hpp"
+#include "rover_msgs/srv/ddb_control.hpp"
+
+class DDBControlNode : public rclcpp::Node
+{
+    static constexpr size_t MAX_CHANNELS = 4;
```
Pmichaud29: 4UL

rover_auxiliary/src/DDB/ddb_control.cpp
```diff
@@ -0,0 +1,373 @@
...
+            {
+                RCLCPP_FATAL(this->get_logger(), "NULL request or response received.");
+                return;
+            }
+            this->ddbControlBank1(*request_, *response_);
+        });
+}
+
+/**
+ * @brief Decides what to do depending on the user requested command for bank 0
```
Pmichaud29: Maybe something more like: Tries to applies user requested command for bank 0

rover_auxiliary/src/DDB/ddb_control.cpp
```diff
@@ -0,0 +1,373 @@
...
+ */
+void DDBControlNode::ddbControlBank0(const rover_msgs::srv::DDBControl::Request& request_,
+                                     rover_msgs::srv::DDBControl::Response& response_)
+{
+    if (request_.channel_id >= MAX_CHANNELS)
+    {
+        RCLCPP_WARN(this->get_logger(), "Invalid channel (0 to 3). Received channel: %d", request_.channel_id);
+        response_.success = false;
+        response_.status = "Invalid channel (0 to 3). Received channel: " + std::to_string(request_.channel_id);
+        return;
```
Pmichaud29: Prefer doing this instead to remove a bit of copy paste

```suggestion
    if (request_.channel_id >= MAX_CHANNELS)
    {
        std::string msg = "Invalid channel, range is [0; " + std::to_string(MAX_CHANNELS - 1UL) + "]. Received channel : " + std::to_string(request_.channel_id);
        
        RCLCPP_WARN(this->get_logger(), msg);
        response_.success = false;
        response_.status = msg;
        return;
```

rover_auxiliary/src/DDB/ddb_control.cpp
```diff
@@ -0,0 +1,373 @@
...
+/**
+ * @brief Changes the state to ON or OFF based on the current state.
+ *
+ * @param channelID_ Received from the user service call
+ * @return true if successfully changed the state of the desired channel else
+ * @return false
+ */
+bool DDBControlNode::toggleChannel(uint8_t channelID_)
+{
+    eToggleState wantedState;
```
Pmichaud29: generally better to set a value even if it's going to be overriden when declaring

rover_auxiliary/src/DDB/ddb_control.cpp
```diff
@@ -0,0 +1,373 @@
...
+    return isUpdated;
+}
+
+/**
+ * @brief Takes the current mode and changes it for a std::string
+ *
+ * @param mode_ Current mode of the channel
+ * @return std::string corresponding to the mode asked
+ */
+std::string DDBControlNode::toStr(eToggleMode mode_)
```
Pmichaud29: suggestion: eToggleModeToStr()

rover_auxiliary/src/DDB/ddb_control.cpp
```diff
@@ -0,0 +1,373 @@
...
+}
+
+/**
+ * @brief Main logic for switching state of the desired channel
+ *
+ * @param request_
+ * @param response_
+ */
+void DDBControlNode::stateLogic(const rover_msgs::srv::DDBControl::Request& request_,
+                                rover_msgs::srv::DDBControl::Response& response_)
```
Pmichaud29: Naming is a bit confusing, setStateLogic or something like this would be a better name

rover_auxiliary/src/DDB/ddb_control.cpp
```diff
@@ -0,0 +1,373 @@
...
+}
+
+/**
+ * @brief Changes the state to ON or OFF based on the current state.
+ *
+ * @param channelID_ Received from the user service call
+ * @return true if successfully changed the state of the desired channel else
+ * @return false
+ */
+bool DDBControlNode::toggleChannel(uint8_t channelID_)
```
Pmichaud29: Toggling here will be hard to sync between this node, the ui and the device. It would be better if the UI request a specific OutputState (On / Off) and you apply it regardless of the current state. The ui should handle the toggling logic with the info it gets from the device

rover_auxiliary/src/DDB/ddb_control.cpp
```diff
@@ -0,0 +1,373 @@
...
+}
+
+/**
+ * @brief Changes the mode to FIX or PWM based on the current mode.
+ *
+ * @param channelID_ Received from the user service call
+ * @return true
+ * @return false
+ */
+bool DDBControlNode::toggleMode(uint8_t channelID_)
```
Pmichaud29: Here also use a setOutputMode instead of a toggle

rover_auxiliary/src/DDB/ddb_control.cpp
```diff
@@ -0,0 +1,373 @@
...
+/**
+ * @brief Changes the values for the PWM mode when requested.
+ *
+ * @param dutyCycle_ Value between 0 and 100 received from user service call
+ * @param frequency_ Value higher than 0 received from user service call
+ * @param channelID_ Received from user service call
+ * @return true
+ * @return false
+ */
+bool DDBControlNode::modifyPWM(uint8_t dutyCycle_, float frequency_, uint8_t channelID_)
```
Pmichaud29: Here no need to check current values, just apply the data you received. If its the same nothing will happen and that's ok

rover_auxiliary/src/DDB/ddb_control.cpp
```diff
@@ -0,0 +1,373 @@
...
+        response_.success = false;
+        response_.status = "Mode could not be toggled for channel #" + std::to_string(request_.channel_id);
+    }
+    else
+    {
+        RCLCPP_INFO(this->get_logger(), "Successfully toggled mode for channel #%d", request_.channel_id);
+        response_.success = true;
+        response_.status = "Successfully toggled mode for channel #" + std::to_string(request_.channel_id);
+    }
+}
```
Pmichaud29: Naming here too

rover_msgs/srv/DDBControl.srv
```diff
@@ -0,0 +1,13 @@
...
+
+# Request
+uint8 channel_id
+uint8 command
+float32 frequency
+uint8 duty_cycle
+---
+# Response
+bool success
+string status
```
Pmichaud29: I would prefer if you try to approch it more like this:
This will change a bit the logic of the services

```suggestion
# output_state
uint8 OUTPUT_OFF
uint8 OUTPUT_ON
uint8 OUTPUT_PWM

# Request
uint8 outputState 
uint8 channelId
float32 frequency # Range is ]0.0; 100.0]  
float32 dutyCycle # Range is [0.0; 100.0]  
---
# Response
bool success
uint8 current_output_ # Value in outputState
string status
```

Pmichaud29 (changes requested):
Peux tu aussi enlever le noeud "light_control" et "compass_calibrator" ? C'est pu d'actualité

Mostly revise the vocabulary used. Right now it's a bit confusing. Think of it more in a "Output" way than a "Toggling" way 

Check the suggestion for the .srv definition first and go from there.

Otherwise the node and code organisation is great.

rover_auxiliary/src/DDB/ddb_control_node.hpp
```diff
@@ -0,0 +1,58 @@
+#ifndef __DDB__NODE__HPP__
+#define __DDB__NODE__HPP__
```
Pmichaud29: DDB_NODE_HPP

rover_auxiliary/src/DDB/ddb_control_node.hpp
```diff
@@ -0,0 +1,58 @@
...
+    {
+        FIX,
+        PWM,
+    };
+
+    struct sChannelInfo
+    {
+        eOutputState state = eOutputState::ON;
+        eOutputMode mode = eOutputMode::FIX;
+        uint8_t dutyCycle = 0;
```
Pmichaud29: We'll want float value for duty cycle

rover_auxiliary/src/DDB/ddb_control_node.hpp
```diff
@@ -0,0 +1,58 @@
...
+    ~DDBControlNode() = default;
+
+  private:
+    void ddbControlBank0(const rover_msgs::srv::DDBControl::Request& request_, rover_msgs::srv::DDBControl::Response& response_);
+    void ddbControlBank1(const rover_msgs::srv::DDBControl::Request& request_, rover_msgs::srv::DDBControl::Response& response_);
+    void setModeLogic(const rover_msgs::srv::DDBControl::Request& request_, rover_msgs::srv::DDBControl::Response& response_);
+    void setStateLogic(const rover_msgs::srv::DDBControl::Request& request_, rover_msgs::srv::DDBControl::Response& response_);
+    void setValuesLogic(const rover_msgs::srv::DDBControl::Request& request_, rover_msgs::srv::DDBControl::Response& response_);
+
+    bool setChannelOutput(uint8_t channelID_, uint8_t desiredState_);
```
Pmichaud29: desiredState should be of type eOutputState instead of uint8_t

rover_auxiliary/src/DDB/ddb_control.cpp
```diff
@@ -0,0 +1,417 @@
...
+
+/**
+ * @brief Tries to apply the user requested command for bank 0
+ *
+ * @param request_
+ * @param response_
+ */
+void DDBControlNode::ddbControlBank0(const rover_msgs::srv::DDBControl::Request& request_,
+                                     rover_msgs::srv::DDBControl::Response& response_)
```
Pmichaud29: The name of the function should explicit that this is a callback

rover_auxiliary/src/DDB/ddb_control.cpp
```diff
@@ -0,0 +1,417 @@
...
+                          + "]. Received channel : " + std::to_string(request_.channel_id);
+        RCLCPP_WARN(this->get_logger(), msg.c_str());
+        response_.success = false;
+        response_.status = msg;
+        return;
+    }
+
+    switch (request_.output_state)
+    {
+        case rover_msgs::srv::DDBControl::Request::OUTPUT_ON:
```
Pmichaud29, line 67: Add the fallthrough flag when intented 
https://en.cppreference.com/w/cpp/language/attributes/fallthrough

rover_auxiliary/src/DDB/ddb_control.cpp
```diff
@@ -0,0 +1,417 @@
...
+            }
+            else
+            {
+                std::string msg = "Current state is OFF. Cannot set PWM mode.";
+                RCLCPP_ERROR(this->get_logger(), msg.c_str());
+                response_.success = false;
+                response_.current_output_state = eOutputStateToUint8_t(_channelInfo[request_.channel_id].state);
+                response_.status = msg;
+                return;
+            }
```
Pmichaud29: Going from OFF to PWM should be a valid command

rover_auxiliary/src/DDB/ddb_control.cpp
```diff
@@ -0,0 +1,417 @@
...
+ * @param response_
+ * @return true if both checks are valid else
+ * @return false if either check isn't valid
+ */
+bool DDBControlNode::valuesCheck(float dutyCycle_,
+                                 float frequency_,
+                                 uint8_t channelID_,
+                                 rover_msgs::srv::DDBControl::Response& response_)
+{
+    if (dutyCycle_ > 100)
```
Pmichaud29: Missing lower bondary and comparing float with int

rover_auxiliary/src/DDB/ddb_control.cpp
```diff
@@ -0,0 +1,417 @@
...
+    {
+        std::string msg = "Duty cycle must be in range [0.0; 100.0]. Duty cycle received: " + std::to_string(dutyCycle_);
+        RCLCPP_ERROR(this->get_logger(), msg.c_str());
+        response_.success = false;
+        response_.current_output_state = eOutputStateToUint8_t(_channelInfo[channelID_].state);
```
Pmichaud29: use TO_UNDERLYING() in rovus_lib instead

rover_auxiliary/src/DDB/ddb_control.cpp
```diff
@@ -0,0 +1,417 @@
...
+        response_.status = msg;
+        return false;
+    }
+
+    if (frequency_ == 0.0)
```
Pmichaud29: 0.0F

and missing check against negative values

rover_auxiliary/src/DDB/ddb_control.cpp
```diff
@@ -0,0 +1,417 @@
...
+    {
+        wantedState = eOutputState::ON;
+    }
+    else if (desiredState_ == rover_msgs::srv::DDBControl::Request::OUTPUT_OFF)
+    {
+        wantedState = eOutputState::OFF;
+    }
+    else
+    {
+        RCLCPP_ERROR(this->get_logger(), "Invalid state for channel: %d", channelID_);
```
Pmichaud29: Define your enum like this:

enum class eOutputState
{
    ON = rover_msgs::srv::DDBControl::Request::OUTPUT_ON
    OFF = rover_msgs::srv::DDBControl::Request::OUTPUT_OFF
};

This way you can then do: 
wantedState = TO_UNDERLYING(desiredState_);

It will make it simpler to use and will prevent possible mapping errors

rover_auxiliary/src/DDB/ddb_control.cpp
```diff
@@ -0,0 +1,417 @@
...
+/**
+ * @brief Changes the values for the PWM mode when requested.
+ *
+ * @param dutyCycle_ Value between 0 and 100 received from user service call
+ * @param frequency_ Value higher than 0 received from user service call
+ * @param channelID_ Received from user service call
+ * @return true
+ * @return false
+ */
+bool DDBControlNode::modifyPWM(float dutyCycle_, float frequency_, uint8_t channelID_)
```
Pmichaud29: Use set/get method naming pls

rover_auxiliary/src/DDB/ddb_control.cpp
```diff
@@ -0,0 +1,417 @@
...
+    switch (mode_)
+    {
+        case eOutputMode::FIX:
+            return "FIX";
+        case eOutputMode::PWM:
+            return "PWM";
+        default:
+            return "UNKNOWN";
+    }
+}
```
Pmichaud29: This expose the internal working of the service but in the service call there's no mention of FIX or PWM. This will be confusing for users

rover_auxiliary/src/DDB/ddb_control_node.hpp
```diff
@@ -0,0 +1,58 @@
...
+    {
+        OFF = 0,
+        ON = 1,
+    };
+
+    enum class eOutputMode : size_t
+    {
+        FIX,
+        PWM,
+    };
```
Pmichaud29: Still unsure about splitting State and Mode, State could be:

enum class eOutputState: uint8_t 
{
  OFF = ...::OUTPUT_OFF,
  PWM = ...::OUTPUT_PWM,
  ON = ...::OUTPUT_ON
}

which would make the rest of the logic a lot simpler

Pmichaud29 (changes requested):

Pmichaud29 (approved):

PR #117, opened 2025-04-25
Small changes to VideoPlayer and CameraInfo

Pmichaud29:
Switch to icons in videoplayer:
- It's uglier but allows more space for the camera url, I added a task in project to generate/make better looking icons for UI.

Fixed invalid name crash in default cameras videoplayer.
- If the CameraInfo map was changed the videoplayermanager would crash at construction, i've added a check before construction so now it constructs with defaults value instead of crashing if the camera name is invalid.

Also reoganized a bit the CameraInfo Namespace

### DIFF
--- a/rover_msgs/include/rovus_lib/camera_info.cpp
+++ b/rover_msgs/include/rovus_lib/camera_info.cpp
@@ -4,11 +4,14 @@ namespace CameraInfo
 {
     bool getNameFromURL(const std::string& url_, std::string& rName_)
     {
-        std::map<std::string, std::string> cameraNameMap;
-        for (const auto& [key, value] : CAMERA_URL_MAP)
-        {
-            cameraNameMap[value] = key;
-        }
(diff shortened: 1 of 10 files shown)


rover_msgs/include/rovus_lib/camera_info.cpp
```diff
@@ -0,0 +1,28 @@
+#include "camera_info.hpp"
+
+namespace CameraInfo
+{
+    bool getNameFromURL(const std::string& url_, std::string& rName_)
```
Gabduf, line 5: good its a lot better than having the 2 maps haha

Gabduf (approved):

rover_msgs/include/rovus_lib/camera_info.cpp
Pmichaud29: yeah hahah

PR #119, opened 2025-04-29
Helpers scripts for canbus debugging

Pmichaud29:


Ndumoulin (approved):
Seems good, but cannot test since I don't have a USB to CAN adapter right now

PR #120, opened 2025-04-29
[GUI] Switched to current user instead of hardcoded phil user in filetransfer widget

Pmichaud29:


Gabduf (approved):
Good and it worked well for me

PR #121, opened 2025-04-29
[GUI] Fixed GUI not exiting on window closed + Added Ctrl+W exit + Made gui exit on any window closed

Pmichaud29:


rover_gui/src/MainWindow.cpp
```diff
@@ -1,12 +1,15 @@
...
+#include "Global/Constant/Keybinding.hpp"
+
 #include <QStackedWidget>
 
 MainWindow::MainWindow(std::shared_ptr<rclcpp::Node> guiNode_):
     QMainWindow(nullptr),
     _centralWidget(this),
     _layout(&_centralWidget),
     _stackedWidget(this),
+    _closeShortCut(Constants::Keybinding::CLOSE_APP, this),
```
Gabduf, line 12: very tiny detail but as copilot said shortcut is one word. However i agree it's easier to read as _closeShortCut

Gabduf (approved):

rover_gui/src/MainWindow.cpp
Pmichaud29: :')

PR #144, opened 2025-05-15
Variables modif  in teleop.cpp to constants.hpp

P3pitosangria:
Put private constants and members into the constant library.


rover_drive_train/src/teleop.cpp
```diff
@@ -15,84 +15,54 @@ class Teleop : public rclcpp::Node
...
-        this->get_parameter("_smallestRadius", _smallestRadius);
-    }
-
-    void CB_joy(const rover_msgs::msg::Joy::SharedPtr msg)
+    void CB_joy(const rover_msgs::msg::Joy& msg)
     {
         rover_msgs::msg::PropulsionMotor message;
 
-        _deadmanSwitch = msg->joy_data[rover_msgs::msg::Joy::L1];
+        float controlMapFactor = 1.0f - Constants::DriveTrain::SMALLEST_RADIUS;
```
Pmichaud29: [nitpick] Only declare in require scope, for example just before this line:
```cpp
float adjustedFactor;
```

This won't allocate memory when unecessery

rover_drive_train/src/teleop.cpp
```diff
@@ -15,84 +15,54 @@ class Teleop : public rclcpp::Node
...
 
-        _linearInput = msg->joy_data[rover_msgs::msg::Joy::JOYSTICK_LEFT_FRONT];
-        _angularInput = msg->joy_data[rover_msgs::msg::Joy::JOYSTICK_LEFT_SIDE];
-        _modeTankAngularInput = msg->joy_data[rover_msgs::msg::Joy::JOYSTICK_RIGHT_SIDE];
-        _modeNormalEnable = msg->joy_data[rover_msgs::msg::Joy::R1];
-        _modeTurboEnable = msg->joy_data[rover_msgs::msg::Joy::R2];
+        float deadmanSwitch = msg.joy_data[Constants::KeyBinding::DEADMAN_SWITCH];
+        float linearInput = msg.joy_data[Constants::KeyBinding::LINEAR_INPUT];
+        float angularInput = msg.joy_data[Constants::KeyBinding::ANGULAR_INPUT];
```
Pmichaud29: Convert to bools to make condition like if (deadmanSwitch) more safe. ([floating points errors](https://stackoverflow.com/questions/2100490/floating-point-inaccuracy-examples) might make this condition true when unwanted since 0.000001 would make this if valid. Either make a helper function to convert floats to bool or do:
if (deadmanswitch == 1.0F)

rover_drive_train/src/teleop.cpp
```diff
@@ -15,84 +15,54 @@ class Teleop : public rclcpp::Node
...
+            float speedFactor = Constants::DriveTrain::SPEED_FACTOR_NORMAL;
 
-            if (_modeNormalEnable)
+            if (modeNormalEnable)
             {
-                speedFactor = _speedFactorNormal;
+                speedFactor = Constants::DriveTrain::SPEED_FACTOR_NORMAL;
             }
-            if (_modeTurboEnable > 0.5 && _modeNormalEnable)
+            if (modeTurboEnable > 0.5 && modeNormalEnable)
```
Pmichaud29: [nitpick] Use type literral for floats: 0.5F, also use caps (I know you didn't write most of this node but it would be a improvement)

rover_drive_train/src/teleop.cpp
```diff
@@ -113,7 +83,7 @@ class Teleop : public rclcpp::Node
 Teleop::Teleop():
     Node("teleop")
 {
-    this->getParams();
+    // this->getParams();
```
Pmichaud29: remove

rover_msgs/msg/Joy.msg
Pmichaud29, line 1: Thanks lol that was bad

Pmichaud29 (changes requested):
Minor things, great job!

Double check the pull request in rover_micro I believe you've uploaded a old version of the lib and not the same one as in this pull request

rover_shared_libs/rover_lib2/src/rover_lib2/helpers/constants.hpp
```diff
@@ -4,33 +4,57 @@
...
+
+    namespace DriveTrain
+    {
+        constexpr float SPEED_FACTOR_CRAWLER = 0.2f;
+        constexpr float SPEED_FACTOR_NORMAL = 0.5f;
+        constexpr float SPEED_FACTOR_TURBO = 1.0f;
+        constexpr float SMALLEST_RADIUS = 0.3f;
+    }  // namespace DriveTrain
+
+    namespace KeyBinding
```
Pmichaud29: We would want this in drivetrain since arm also has it's keybindings, so I would suggest:
DriveTrain::Keybindings::*

Pmichaud29 (dismissed):
One very small comment good job!

AraaRobot (approved):

PR #148, opened 2025-05-20
Status page for devices on the rover

AraaRobot:
Status page allowing the user to see different devices on the rover
  - Push button that calls a service asking for the error_state of each device
  - Changing color of the devices depending on the error_state received (Green = ok, orange = warning, red = error)
  - Label for each device with its name, its ID (in hex) and how many times it has rebooted
  - Grid containing each device and their informations


Pmichaud29 (commented):
Comments on UI look:
- Might want to order id in order or regroup in sections by ids range. 
- Might want to add an icon for each device to give a faster overview of which is which.
- I've attached a possible avenue of design to give you ideas.
- Regoup all info in the same "block" this module will probably be on the dashboard and it's allocated space will be minimal

image

rover_gui/src/QDeviceStatus/QDeviceStatus.cpp
```diff
@@ -0,0 +1,251 @@
...
+    _deviceReboot[TO_UNDERLYING(RoverCan2::Constant::eDeviceId::CAMERA_ROVER_MAIN)] = _ui.cameraMainInfo;
+    _deviceReboot[TO_UNDERLYING(RoverCan2::Constant::eDeviceId::CAMERA_ROVER_ANTENNA)] = _ui.cameraAntenneInfo;
+    _deviceReboot[TO_UNDERLYING(RoverCan2::Constant::eDeviceId::LIGHTS_MAIN)] = _ui.lightsMainInfo;
+    // _deviceReboot[TO_UNDERLYING(RoverCan2::Constant::eDeviceId::SWITCHETH0)] = _ui.switchETH0Info;
+    // _deviceReboot[TO_UNDERLYING(RoverCan2::Constant::eDeviceId::SWITCHETH1)] = _ui.switchETH1Info;
+
+    _sub_deviceStatus = _node->create_subscription<rover_msgs::msg::CanDeviceStatus>(
+        "/rover/can/devices_status",
+        QOS_DEFAULT,
+        [this](const rover_msgs::msg::CanDeviceStatus::SharedPtr msg)
```
Pmichaud29: Take as const ref instead of shared_ptr

rover_gui/src/QDeviceStatus/QDeviceStatus.cpp
```diff
@@ -0,0 +1,251 @@
...
+    _QStatusWorker->requestDeviceStatusManager(_client_requestErrorStatus);
+}
+
+/**
+ * @brief Handles the response from the service call
+ *
+ * @param success_
+ * @param response_
+ */
+void QDeviceStatus::onRequestDeviceStatusSuccessful(bool success_, const std::string response_)
```
Pmichaud29: take string a ref to avoid unnecessary copy

rover_gui/src/QDeviceStatus/QDeviceStatus.cpp
```diff
@@ -0,0 +1,251 @@
...
+ * @brief Keeps track of the number of reboots for each device. Logs a message when a device reboots.
+ *
+ * @param deviceID_
+ */
+void QDeviceStatus::updateRebootCounter(uint16_t deviceID_)
+{
+    if (!_deviceInfo.contains(deviceID_) || !_deviceReboot.contains(deviceID_))
+    {
+        return;
+    }
```
Pmichaud29: Once this is true, I suggest you to create a temporary variable holding a reference to _numberOfDeviceReboots[deviceID_]. This way you don't have to use the [] operator each time and makes it safer to maintain (because it's easy to miss one)

rover_gui/src/QDeviceStatus/QDeviceStatus.cpp
```diff
@@ -0,0 +1,251 @@
...
+
+    _numberOfDeviceReboots[deviceID_]
+        = _numberOfDeviceRebootsFromButton[deviceID_] + _numberOfCalls - _deviceMessageCount[deviceID_];
```
Pmichaud29: Do this overflows if _deviceMessageCount is higher than `_numberOfDeviceRebootsFromButton[deviceID_] + _numberOfCalls` ? Might be safer to cap at 0 if `_numberOfDeviceRebootsFromButton[deviceID_] + _numberOfCalls - _deviceMessageCount[deviceID_]` is negative

rover_gui/src/QDeviceStatus/QDeviceStatus.cpp
```diff
@@ -0,0 +1,251 @@
...
+ * @param deviceID_
+ */
+void QDeviceStatus::setStatusReport(uint16_t deviceID_)
+{
+    if (!_deviceInfo.contains(deviceID_) || !_deviceReboot.contains(deviceID_))
+    {
+        return;
+    }
+
+    auto label = _deviceReboot[deviceID_];
```
Pmichaud29: CG: Don't use auto unless type is explicited by rvalue

rover_gui/src/QDeviceStatus/QDeviceStatus.cpp
```diff
@@ -0,0 +1,251 @@
...
+ * @brief Sets the default style for all device labels.
+ *
+ */
+void QDeviceStatus::setDefaultStyle()
+{
+    for (auto it = _deviceInfo.begin(); it != _deviceInfo.end(); ++it)
+    {
+        it.value()->setStyleSheet(STATUS_DEFAULT);
+    }
+}
```
Pmichaud29: Might want to use the modern for loop form:
for (auto it& : _deviceInfo)

rover_gui/src/QDeviceStatus/QDeviceStatus.cpp
```diff
@@ -0,0 +1,251 @@
...
+
+const std::string QDeviceStatus::getDeviceName(uint16_t deviceID_)
```
Pmichaud29: No need for this to be const

rover_gui/src/QDeviceStatus/QDeviceStatus.hpp
```diff
@@ -0,0 +1,63 @@
+#ifndef __QDEVICESTATUS_HPP__
+#define __QDEVICESTATUS_HPP__
```
Pmichaud29: include guards new standard will be:
INCLUDE_PATH_H

rover_gui/src/QDeviceStatus/QDeviceStatus.hpp
```diff
@@ -0,0 +1,63 @@
...
+
+// ROS
+#include "rclcpp/rclcpp.hpp"
+#include "rover_msgs/msg/can_device_status.hpp"
+#include "rover_msgs/srv/empty.hpp"
+#include "rover_can2/constant.hpp"
+#include "rover_lib2/helpers/macros.hpp"
+#include "rover_lib2/helpers/constants.hpp"
```
Pmichaud29: include <>

rover_gui/src/QDeviceStatus/QDeviceStatus.cpp
```diff
@@ -0,0 +1,251 @@
...
+constexpr const char* STATUS_ERROR = "QWidget {"
+                                     "background-color : #e57373;"
+                                     "color: black;"
+                                     "border-radius: 5px;"
+                                     "padding: 5px 10px;"
+                                     "}";
+
+QDeviceStatus::QDeviceStatus(std::shared_ptr<rclcpp::Node> guiNode_, QWidget* parent_):
+    QWidget(parent_),
+    _node(guiNode_)
```
Pmichaud29: Assert if node invalid

rover_gui/src/QDeviceStatus/QDeviceStatus.hpp
```diff
@@ -0,0 +1,63 @@
...
+    QLabel* _imageLabel;
+    QPixmap _pixmap;
+
+    rclcpp::Subscription<rover_msgs::msg::CanDeviceStatus>::SharedPtr _sub_deviceStatus;
+    rclcpp::Client<rover_msgs::srv::Empty>::SharedPtr _client_requestErrorStatus;
+
+    std::unordered_map<uint16_t, int16_t> _deviceMessageCount;
+    std::unordered_map<uint16_t, uint16_t> _numberOfDeviceReboots;
+    std::unordered_map<uint16_t, uint16_t> _oldDeviceReboots;
+    std::unordered_map<uint16_t, uint16_t> _numberOfDeviceRebootsFromButton;
```
Pmichaud29: Why not use rover_can2 eDeviceId enum instead?

rover_gui/src/QDeviceStatus/QDeviceStatus.hpp
```diff
@@ -0,0 +1,63 @@
...
+
+    QMap<uint16_t, QWidget*> _deviceInfo;
+    QMap<uint16_t, QLabel*> _deviceReboot;
```
Pmichaud29: Don't use QObjects if std offers an alternative (unordered map / map)

rover_gui/src/QDeviceStatus/QDeviceStatus.hpp
```diff
@@ -0,0 +1,63 @@
...
+
+    uint16_t _numberOfCalls = 0U;
```
Pmichaud29, line 59: might be easier to not track the number of call but just to reset all device reboot counter to 0 before each call. (It's ok to lose information about reboot between each call since they are manual)

rover_gui/src/QDeviceStatus/QDeviceStatus.hpp
```diff
@@ -0,0 +1,63 @@
...
+
+    std::shared_ptr<QStatusWorker> _QStatusWorker;
```
Pmichaud29: Why dynamically allocated? If pointer is absolutely necessary (I don't think it is) use unique_ptr when possible.

rover_gui/src/QDeviceStatus/Worker/QStatusWorker.hpp
```diff
@@ -0,0 +1,24 @@
+#ifndef __QSTATUS_WORKER_HPP__
+#define __QSTATUS_WORKER_HPP__
+
+#include "Global/Workers/QWorker.hpp"
+#include "rclcpp/rclcpp.hpp"
+#include "rover_msgs/srv/empty.hpp"
+#include "rover_lib2/helpers/macros.hpp"
```
Pmichaud29: <>

rover_gui/src/QDeviceStatus/Worker/QStatusWorker.hpp
```diff
@@ -0,0 +1,24 @@
...
+
+  public:
+    QStatusWorker(bool start_ = false, QObject* parent_ = nullptr);
+    ~QStatusWorker();
+
+    void requestDeviceStatusManager(std::shared_ptr<rclcpp::Client<rover_msgs::srv::Empty>> client_requestErrorStatus_);
+    void requestDeviceStatusInternal(std::shared_ptr<rclcpp::Client<rover_msgs::srv::Empty>> client_requestErrorStatus_);
+
+  signals:
+    void onRequestDeviceStatusSuccessful(bool success_, const std::string response_);
```
Pmichaud29: Might want to take response_ as const std::string& to prevent expensive copy

rover_gui/src/QDeviceStatus/Worker/QStatusWorker.cpp
```diff
@@ -0,0 +1,52 @@
+#include "QStatusWorker.hpp"
+
+QStatusWorker::QStatusWorker(bool start_, QObject* parent_):
+    QWorker(start_, parent_)
+{
+}
+
+QStatusWorker::~QStatusWorker()
+{
+    this->finish();
```
Pmichaud29: Isn't this already called by QWorker's destructor?

Pmichaud29 (changes requested):
Multiple small things, shouldn't be changes

rover_gui/src/QDeviceStatus/QDeviceStatus.hpp
```diff
@@ -0,0 +1,63 @@
...
+
+#include <unordered_map>
+
+class QDeviceStatus : public QWidget
+{
+    Q_OBJECT
+
+  public:
+    QDeviceStatus(std::shared_ptr<rclcpp::Node> guiNode_, QWidget* parent_);
+    ~QDeviceStatus() = default;
```
Pmichaud29: Don't specify if default and not virtual

rover_gui/src/QDeviceStatus/QDeviceStatus.hpp
```diff
@@ -0,0 +1,63 @@
...
+    rclcpp::Subscription<rover_msgs::msg::CanDeviceStatus>::SharedPtr _sub_deviceStatus;
+    rclcpp::Client<rover_msgs::srv::Empty>::SharedPtr _client_requestErrorStatus;
+
+    std::unordered_map<RoverCan2::Constant::eDeviceId, uint16_t> _deviceMessageCount;
+    std::unordered_map<RoverCan2::Constant::eDeviceId, uint16_t> _numberOfDeviceReboots;
+    std::unordered_map<RoverCan2::Constant::eDeviceId, uint16_t> _oldDeviceReboots;
+    std::unordered_map<RoverCan2::Constant::eDeviceId, uint16_t> _numberOfDeviceRebootsFromButton;
+
+    std::unordered_map<RoverCan2::Constant::eDeviceId, QWidget*> _deviceInfo;
+    std::unordered_map<RoverCan2::Constant::eDeviceId, QLabel*> _deviceReboot;
```
Pmichaud29: I would conbine all of those in a struct so you only have one map.

something like this:

```cpp
struct sCanDeviceInfos
{
  uint16_t deviceMessageCount;
  uint16_t numberOfDeviceReboots;
  uint16_t oldDeviceReboots;
  uint16_t numberOfDeviceRebootsFromButton;
  QWidget* deviceInfo;
  QLabel* deviceReboot;
};

std::unordered_map<RoverCan2::Constant::eDeviceId, sCanDeviceInfos> _canDevices;
```

It will make it easier to follow in the cpp file, also since they're all necessary for your code to work, it's generally a cleaner practice to combine them

rover_gui/src/QDeviceStatus/QDeviceStatus.cpp
```diff
@@ -0,0 +1,270 @@
...
+{
+    if (!_deviceInfo.contains(deviceID_) || !_deviceReboot.contains(deviceID_))
+    {
+        return;
+    }
+
+    uint16_t& deviceReboots = _numberOfDeviceReboots[deviceID_];
+    uint16_t& oldDeviceReboots = _oldDeviceReboots[deviceID_];
+    uint16_t& deviceRebootsFromButton = _numberOfDeviceRebootsFromButton[deviceID_];
+    uint16_t& deviceMessageCount = _deviceMessageCount[deviceID_];
```
Pmichaud29: could be const

rover_gui/src/QDeviceStatus/QDeviceStatus.cpp
```diff
@@ -0,0 +1,270 @@
...
+        deviceReboots = 0;
+        return;
+    }
+
+    deviceReboots = deviceRebootsFromButton + _numberOfCalls - deviceMessageCount;
+
+    if (deviceReboots != oldDeviceReboots)
+    {
+        oldDeviceReboots = deviceReboots;
+        uint16_t deviceID = static_cast<uint16_t>(deviceID_);
```
Pmichaud29: use std::to_underlying instead

rover_gui/src/QDeviceStatus/QDeviceStatus.cpp
```diff
@@ -0,0 +1,270 @@
...
+{
+    if (!_deviceInfo.contains(deviceID_) || !_deviceReboot.contains(deviceID_))
+    {
+        return;
+    }
+
+    QLabel* label = _deviceReboot[deviceID_];
+
+    std::string deviceName = this->getDeviceName(deviceID_);
+    uint16_t deviceID = static_cast<uint16_t>(deviceID_);
```
Pmichaud29: use std::to_underlying instead

rover_gui/src/QDeviceStatus/QDeviceStatus.cpp
```diff
@@ -0,0 +1,270 @@
...
+ * @brief Sets the default style for all device labels.
+ *
+ */
+void QDeviceStatus::setDefaultStyle()
+{
+    for (auto& it : _deviceInfo)
+    {
+        it.second->setStyleSheet(STATUS_DEFAULT);
+    }
+}
```
Pmichaud29: nitpick: for cpp23 it's prefered to use structured binding declaration:


```suggestion
    for (auto& [key, value] : _deviceInfo)
    {
        value->setStyleSheet(STATUS_DEFAULT);
    }
}
```

rover_gui/src/QDeviceStatus/QDeviceStatus.cpp
```diff
@@ -0,0 +1,270 @@
...
+
+std::string QDeviceStatus::getDeviceName(RoverCan2::Constant::eDeviceId deviceID_)
```
Pmichaud29: func can be declared const

rover_gui/src/QDeviceStatus/Worker/QStatusWorker.hpp
```diff
@@ -0,0 +1,23 @@
+#ifndef QDEVICESTATUS_WORKER_QSTATUS_WORKER_HPP
+#define QDEVICESTATUS_WORKER_QSTATUS_WORKER_HPP
+
+#include <Global/Workers/QWorker.hpp>
```
Pmichaud29:
```suggestion
#include "Global/Workers/QWorker.hpp"
```

rover_gui/src/QDeviceStatus/QDeviceStatus.hpp
```diff
@@ -0,0 +1,63 @@
+#ifndef QDEVICESTATUS_QDEVICESTATUS_HPP
+#define QDEVICESTATUS_QDEVICESTATUS_HPP
+
+// ROS
```
Pmichaud29, line 4:
```suggestion
#include <QWidget>
```

rover_gui/src/QDeviceStatus/QDeviceStatus.cpp
```diff
@@ -0,0 +1,270 @@
+#include "QDeviceStatus.hpp"
+#include <QStyle>
```
Pmichaud29:
```suggestion
#include "QDeviceStatus.hpp"

#include <QStyle>
```

Pmichaud29 (changes requested):
2 things:
 - On my end the reboot counter doesn't seem to work anymore (always stays at 0)
 - On a style note only the camera icon is white while all other icons are black (perso I prefer the white version) but they should all be the same (black or white, your choice)

rover_gui/src/QDeviceStatus/QDeviceStatus.cpp

rover_gui/src/QDeviceStatus/QDeviceStatus.cpp

rover_gui/src/QDeviceStatus/QDeviceStatus.cpp

rover_gui/src/QDeviceStatus/QDeviceStatus.cpp
AraaRobot: Done

rover_gui/src/QDeviceStatus/QDeviceStatus.hpp
AraaRobot: done

rover_gui/src/QDeviceStatus/QDeviceStatus.hpp
AraaRobot: done

rover_gui/src/QDeviceStatus/QDeviceStatus.hpp
AraaRobot: done

rover_gui/src/QDeviceStatus/QDeviceStatus.cpp

rover_gui/src/QDeviceStatus/QDeviceStatus.cpp

rover_gui/src/QDeviceStatus/Worker/QStatusWorker.hpp
AraaRobot: done

rover_gui/src/QDeviceStatus/QDeviceStatus.hpp
```diff
@@ -0,0 +1,66 @@
...
+    void updateDeviceInfo();
+    void updateDeviceColor(RoverCan2::Constant::eDeviceId deviceID_, const rover_msgs::msg::CanDeviceStatus& msg_);
+    void setStatusReport(RoverCan2::Constant::eDeviceId id_);
+    void updateRebootCounter(RoverCan2::Constant::eDeviceId deviceID_);
+    void setDefaultStyle();
+    const std::string getDeviceName(RoverCan2::Constant::eDeviceId deviceID_);
+
+  private slots:
+
+    void onRequestDeviceStatusSuccessful(bool success_, const std::string& response_);
```
Pmichaud29:
```suggestion
  private slots:
    void onRequestDeviceStatusSuccessful(bool success_, const std::string& response_);
```

rover_gui/src/QDeviceStatus/QDeviceStatus.cpp
Pmichaud29: I noticed you set some but some are still static_cast

rover_gui/src/QDeviceStatus/QDeviceStatus.hpp
Pmichaud29: This is a mistake on my part, the QWidget include isn't necessary and the // ROS comment was nice in this context. I've reverted the change to it's initial state

rover_gui/src/QDeviceStatus/QDeviceStatus.cpp
Pmichaud29: The function can be const but the return type shoulnd't as having a const std::string is mean to the user because he'll need to make copy of the string even if you already return a copy

std::string QDeviceStatus::getDeviceName(RoverCan2::Constant::eDeviceId deviceID_) const; would be the correct signature

Pmichaud29 (commented):
I've push some the last smal comment I had for you to be able to merge directly but I noticed the icon for the video player don't work anymore only in this branch. Might want to look into this

Pmichaud29 (approved):
Good job!

PR #174, opened 2025-06-14
Error fix for nav map

emilesavoie:


Pmichaud29:
Closing because I believe this was just to check a diff

PR #180, opened 2025-06-19
QDeviceStatus Widget changes

AraaRobot:
- Widget is now reinitialised before service call
- Counter effectively goes up when more than 1 message are received after a request
- Color priority respected (ok = green if not rebooted, rebooted = yellow, warn and error = red)
- Cleaning old code that wasnt being used


rover_gui/src/QDeviceStatus/QDeviceStatus.cpp
```diff
@@ -202,15 +191,18 @@ void QDeviceStatus::setStatusReport(RoverCan2::Constant::eDeviceId deviceID_)
 void QDeviceStatus::updateDeviceColor(RoverCan2::Constant::eDeviceId deviceID_, const rover_msgs::msg::CanDeviceStatus& msg_)
 {
     QWidget* widgetInfo = _canDevices[deviceID_].deviceInfo;
+    QString currentStyle = widgetInfo->styleSheet();
```
Pmichaud29: If possible take as a const & to avoid a big copy

rover_gui/src/QDeviceStatus/QDeviceStatus.cpp
```diff
@@ -202,15 +191,18 @@ void QDeviceStatus::setStatusReport(RoverCan2::Constant::eDeviceId deviceID_)
...
 
     switch (msg_.error_state)
     {
         case rover_msgs::msg::CanDeviceStatus::STATUS_OK:
-            widgetInfo->setStyleSheet(STATUS_SUCCESS);
+            if (currentStyle != STATUS_WARNING)
```
Pmichaud29: I think we should also check if STATUS_ERROR to not overwrite errors too

rover_gui/src/QDeviceStatus/QDeviceStatus.cpp

rover_gui/src/QDeviceStatus/QDeviceStatus.cpp

Pmichaud29 (approved):

PR #184, opened 2025-07-16
Instauring QFlowLayout into QDeviceStatus

AraaRobot:
The QDeviceStatus widget now uses a flow layout that rearranges the internal widgets horizontally depending on the allowed space.

- Stretching ratio in dashboard of 6(For navigation widget column):1(device + arbitration widgets column)

- Fixed information widgets size whenever the normal QDeviceStatus page is on

- Only the icons are shown in the dashboard


rover_gui/src/Global/QFlowLayout.cpp
```diff
@@ -0,0 +1,162 @@
...
+    m_vSpace(vSpacing)
+{
+    setContentsMargins(margin, margin, margin, margin);
+}
+
+QFlowLayout::~QFlowLayout()
+{
+    QLayoutItem* item;
+    while ((item = takeAt(0)))
+        delete item;
```
Pmichaud29, line 29: deletelater perhaps? to let qt garbage collector handle it?

rover_gui/src/Global/QFlowLayout.cpp
```diff
@@ -0,0 +1,162 @@
...
+    else
+    {
+        return smartSpacing(QStyle::PM_LayoutVerticalSpacing);
+    }
+}
+
+int QFlowLayout::count() const
+{
+    return itemList.size();
+}
```
Pmichaud29: conversion from long long to int

rover_gui/src/Global/QFlowLayout.cpp
```diff
@@ -0,0 +1,162 @@
...
+        size = size.expandedTo(item->minimumSize());
+
+    const QMargins margins = contentsMargins();
+    size += QSize(margins.left() + margins.right(), margins.top() + margins.bottom());
+    return size;
+}
+
+int QFlowLayout::doLayout(const QRect& rect, bool testOnly) const
+{
+    int left, top, right, bottom;
```
Pmichaud29, line 118: one line per declaration, this is error prone

rover_gui/src/Global/QFlowLayout.cpp
```diff
@@ -0,0 +1,162 @@
...
+    getContentsMargins(&left, &top, &right, &bottom);
+    QRect effectiveRect = rect.adjusted(+left, +top, -right, -bottom);
```
Pmichaud29: explicit "+" not needed

rover_gui/src/Global/QFlowLayout.cpp
```diff
@@ -0,0 +1,162 @@
...
+int QFlowLayout::smartSpacing(QStyle::PixelMetric pm) const
+{
+    QObject* parent = this->parent();
+    if (!parent)
+    {
+        return -1;
+    }
+    else if (parent->isWidgetType())
+    {
+        QWidget* pw = static_cast<QWidget*>(parent);
```
Pmichaud29, line 161: take the parent as a QWidget* directly then otherwise this is a invalid cast

rover_gui/src/Global/QFlowLayout.hpp
```diff
@@ -0,0 +1,36 @@
...
+    QSize minimumSize() const override;
+    void setGeometry(const QRect& rect) override;
+    QSize sizeHint() const override;
+    QLayoutItem* takeAt(int index) override;
+
+  private:
+    int doLayout(const QRect& rect, bool testOnly) const;
+    int smartSpacing(QStyle::PixelMetric pm) const;
+
+    QList<QLayoutItem*> itemList;
```
Pmichaud29: Take items in a std::list and take as a smart_ptr (unique/shared/weak). You don't want to handle the deletion yourself (this hard to track and error prone).

This way should won't need to define a destructor since the std::list destructor will handle it

rover_gui/src/QDeviceStatus/QDeviceStatus.cpp
```diff
@@ -40,17 +39,15 @@ QDeviceStatus::QDeviceStatus(std::shared_ptr<rclcpp::Node> guiNode_, QWidget* pa
...
-    _canDevices[RoverCan2::Constant::eDeviceId::REARLEFT_MOTOR] = {0, 0, _ui.rearleftMotor, _ui.rearleftMotorInfo};
-    _canDevices[RoverCan2::Constant::eDeviceId::REARRIGHT_MOTOR] = {0, 0, _ui.rearrightMotor, _ui.rearrightMotorInfo};
-    _canDevices[RoverCan2::Constant::eDeviceId::GNSS] = {0, 0, _ui.gnss, _ui.gnssInfo};
-    _canDevices[RoverCan2::Constant::eDeviceId::DDB_CONTROLLER] = {0, 0, _ui.ddbController, _ui.ddbControllerInfo};
-    _canDevices[RoverCan2::Constant::eDeviceId::CAMERA_ROVER_MAIN] = {0, 0, _ui.cameraMain, _ui.cameraMainInfo};
-    _canDevices[RoverCan2::Constant::eDeviceId::CAMERA_ROVER_ANTENNA] = {0, 0, _ui.cameraAntenne, _ui.cameraAntenneInfo};
-    _canDevices[RoverCan2::Constant::eDeviceId::LIGHTS_MAIN] = {0, 0, _ui.lightsMain, _ui.lightsMainInfo};
-    // _canDevices[RoverCan2::Constant::eDeviceId::SWITCHETH0] = {0, 0, _ui.switchETH0, _ui.switchETH0Info};
-    // _canDevices[RoverCan2::Constant::eDeviceId::SWITCHETH1] = {0, 0, _ui.switchETH1, _ui.switchETH1Info};
+    _layout = new QFlowLayout(_ui.deviceInfos);
```
Pmichaud29: don't use neither new or delete

Pmichaud29 (changes requested):
redo a check in our coding guideline for variable naming (args missing trailing "_", private member missing "_" prefix ) 

Remove all instance of new and delete and replace them with either stack variable or smart pointers

rover_gui/src/QDeviceStatus/QDeviceStatus.cpp

rover_gui/src/Global/QFlowLayout.cpp
AraaRobot: From Qt example

rover_gui/src/Global/QFlowLayout.cpp
AraaRobot: From Qt example

rover_gui/src/Global/QFlowLayout.cpp
AraaRobot: From Qt example

rover_gui/src/Global/QFlowLayout.cpp
AraaRobot: From Qt example

rover_gui/src/Global/QFlowLayout.cpp
AraaRobot: From Qt example

rover_gui/src/Global/QFlowLayout.hpp
AraaRobot: From Qt example

Pmichaud29 (dismissed):

PR #188, opened 2025-07-20
Face north button on map

AraaRobot:
Added button that allows the user to face North. When pressed, the camera faces North and the button becomes "activated" (similar to tracking button). Whenever the user faces another direction, the button becomes "disabled"

When top view is activated, the camera is changed to look towards North.

Because of Cesium, lock on North is only possible in 2D (Top down view) since the Earth is round and north changes when the user moves the camera to far and we can't deactivate rotation/tilting on only 1 axis


rover_gui/src/Global/Constant/other/map.html
```diff
@@ -213,9 +228,95 @@ <h2>Not able to connect to internet</h2>
...
+					}
+				});
+				isAdjustingCamera = false;
+			}
+		}	
+	}
+
+	function stillFacingNorth() 
+	{
+		if (!isFacingNorth || isAdjustingCamera) return;
```
Pmichaud29:
```suggestion
		if (!isFacingNorth || isAdjustingCamera)
		{ 
			return;
		}
```

Pmichaud29 (dismissed):
Works great, I've notice something when trying it:
- All the buttons seems to hold a state even when they get deselected automatically. For example when the track position is set and the top down view is too, after unselecting the track position you can't move the map unless you also unselect top down view and toggle on/off the track position

rover_gui/src/Global/Constant/other/map.html
AraaRobot: done

Pmichaud29 (approved):

PR #201, opened 2025-08-02
Moving with rover when tracking

AraaRobot:
Each time GUI receives new position and if currently tracking, camera moves with rover

### DIFF
--- a/rover_sim/src/gps_test_pub.cpp
+++ b/rover_sim/src/gps_test_pub.cpp
@@ -29,6 +29,8 @@ class GpsPublisher : public rclcpp::Node
 
         publisher_->publish(msg);
 
+        latitude_ += 0.00001;
+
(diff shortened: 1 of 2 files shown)


rover_sim/src/gps_test_pub.cpp
```diff
@@ -29,6 +29,8 @@ class GpsPublisher : public rclcpp::Node
 
         publisher_->publish(msg);
 
+        latitude_ += 0.00001;
+
```
Pmichaud29: already done on line 40

Pmichaud29 (dismissed):

rover_sim/src/gps_test_pub.cpp
AraaRobot: true, changed it back

Pmichaud29 (approved):

PR #224, opened 2025-08-10
Added heading to GUI when calculating distance of a waypoint from Rover

AraaRobot:
PR #223 must be merge beforehand. This PR uses files from it


emilesavoie (approved):
Youpi

PR #227, opened 2025-08-10
Fixing old path loading

AraaRobot:
Fixed the old path loading to load everything only once and create one single path without removing it.

Current path should not flicker anymore due to constant incoming data.

If map was never online, when it becomes online, it will initiate the map (Only if it wasnt online at startup)

### DIFF
--- a/rover_gui/src/Global/Constant/other/JS/pathManager.js
+++ b/rover_gui/src/Global/Constant/other/JS/pathManager.js
@@ -74,61 +74,62 @@ class PathManager
 
     drawPathTaken(latitude_, longitude_, name_)
     {
-        let newEntity = null;
-        if (name_ === PathManager.OLD_POSITION_PATH)
+        this.#currentPath.push([longitude_, latitude_]);
+        const flatPositions = this.#currentPath.flat();
+
+        const idx = this.#pathEntities.findIndex(e => e.id === PathManager.POSITION_PATH);
+        
+        if (idx !== -1) 
         {
-            this.#oldPath.push([longitude_, latitude_]);
-            const flatPositions = this.#oldPath.flat();
-    
-            const idx = this.#pathEntities.findIndex(e => e.id === PathManager.OLD_POSITION_PATH);
-            
-            if (idx !== -1) 
-            {
-                this.viewer.entities.remove(this.#pathEntities[idx]);
-                this.#pathEntities.splice(idx, 1);
-            }
-    
-            newEntity = this.viewer.entities.add({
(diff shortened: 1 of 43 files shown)


rover_gui/src/Global/Constant/other/JS/pathManager.js
```diff
@@ -74,62 +74,59 @@ class PathManager
...
-            newEntity = this.viewer.entities.add({
+            // Create the entity on the first call
+            this.#currentPathEntity = this.viewer.entities.add({
                 id: name_,
                 polyline: {
                     positions: Cesium.Cartesian3.fromDegreesArray(flatPositions),
                     width: 3,
                     material: new Cesium.PolylineOutlineMaterialProperty({
-                        color: Cesium.Color.CYAN,
+                        color: Cesium.Color.RED,
```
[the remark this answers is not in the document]
AraaRobot: No?

Pmichaud29 (dismissed):

rover_gui/src/QNavigation/QNavigation.cpp
```diff
@@ -139,8 +136,12 @@ void QNavigation::createNavigationFolder(void)
 void QNavigation::onGpsMessage(const rover_msgs::msg::Gps& msg_)
 {
     emit this->gpsCallback(msg_.latitude, msg_.longitude, msg_.heading);
-    emit this->updatePathTaken(msg_.latitude, msg_.longitude, QString::fromStdString(_pathManager.PATH_NAME));
-    _pathManager.writePosToCSV(msg_.latitude, msg_.longitude);
+    emit this->updatePathTaken(msg_.latitude, msg_.longitude);
+
+    if (++_gpsMsgCounter % GPS_SKIP_RATE == 0)
```
emilesavoie: Not sure about this here. This is non critical so not a hastle. Just wondering if another approach could be used suc has logging at a set time interval or at a set distance interval. If you deem this is the cleanest way, just add a small comment in the header file the GPS_SKIP_RATE parameter

emilesavoie (commented):
Everything about the code in itself is good. Just checkout the small comment I made and once the conversation is resolved I will approve

rover_gui/src/QNavigation/QNavigation.cpp
AraaRobot: I agree, and i fixed it

emilesavoie (dismissed):
All good

emilesavoie (approved):

PR #262, opened 2026-01-15
added bash file for ui launch

### LINKED ISSUE #264: Fix UI shutdown on open

emilesavoie:
Added bash file for more robustness on ui startup

### DIFF
--- a/rover_msgs/CMakeLists.txt
+++ b/rover_msgs/CMakeLists.txt
@@ -8,6 +8,7 @@ endif()
 set(INSTALL_PATH "${CMAKE_INSTALL_PREFIX}/..")
 
 set(ROS_LAUNCH_SCRIPT "${CMAKE_CURRENT_SOURCE_DIR}/../scripts/launch_ui.sh")
+set(ICON_NAME "exec_icon.svg")
 set(ROVER_BASE_ICON_PATH_PREFIX "${INSTALL_PATH}/rover_gui/lib/rover_gui")
 set(ROVER_BASE_ICON_PATH "${ROVER_BASE_ICON_PATH_PREFIX}/${ICON_NAME}")
 set(DESKTOP_FILE_INSTALL_PATH "$ENV{HOME}/.local/share/applications/rover_base.desktop")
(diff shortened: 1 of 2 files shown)


rover_msgs/CMakeLists.txt
```diff
@@ -17,7 +18,7 @@ set(DESKTOP_FILE
       "[Desktop Entry]\n"
       "Type=Application\n"
       "Name=Rover Base\n"
-      "Exec=bash -c 'source ~/.bashrc && ${ROVER_BASE_EXEC_PATH}'\n"
+      "Exec=bash -c '${ROS_LAUNCH_SCRIPT}; exec bash'\n"
```
AraaRobot: If the point of the `exec bash` is to keep the terminal open after closing, it doesn't work properly. The default terminal on Ubuntu is set to close on script exit, even with the exec bash. See next comment for idea

rover_msgs/CMakeLists.txt
```diff
@@ -7,6 +7,7 @@ endif()
 
 set(INSTALL_PATH "${CMAKE_INSTALL_PREFIX}/..")
 
+set(ROS_LAUNCH_SCRIPT "${CMAKE_CURRENT_SOURCE_DIR}/../scripts/launch_ui.sh")
 set(ROVER_BASE_EXEC_PATH "${INSTALL_PATH}/rover_gui/lib/rover_gui/main_gui")
```
AraaRobot:
```suggestion

set(ROS_LAUNCH_SCRIPT "${CMAKE_CURRENT_SOURCE_DIR}/../scripts/launch_ui.sh")
```
ROVER_BASE_EXEC_PATH variable is not used anymore, we could delete it

scripts/launch_ui.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+set -e
+source /opt/ros/humble/setup.bash
+source ~/ros2_ws/install/setup.bash
+ros2 launch rover_gui gui.launch.py
```
AraaRobot, line 6:
```suggestion
#!/bin/bash

set -e
source /opt/ros/humble/setup.bash
source ~/ros2_ws/install/setup.bash
ros2 launch rover_gui gui.launch.py
read -p "Press enter to close..."
```
This would allow the terminal to stay open until input and we could remove the previous exec bash

AraaRobot (changes requested):
It works

scripts/launch_ui.sh
emilesavoie: My view is that the terminal should close on exit. Would you want to keep it open for potential logs? @AraaRobot

AraaRobot (approved):

PR #280, opened 2026-02-18
Dev/bms logger jbl

JoeyBilodeauL:


emilesavoie (changes requested):
Before doing a full review, try seperating the reader functionnality in seperate "worker class" like BmsLogger or something. Then you can use an instance of this class within your ros2 node. This allows for encapsulation of ros2 functionnalities (pub/sub/src, etc...) vs custom interfaces. You can take a look at rover_arm for inspiration on how to implement this!

If you have questions or want to look at it together feek free!

PR #288, opened 2026-03-03
Dev/serial com jbl

JoeyBilodeauL:
class for serial com

### DIFF
--- a/rover_shared_libs/rover_lib2/src/rover_lib2/communication/Serial/serial_com.cpp
+++ b/rover_shared_libs/rover_lib2/src/rover_lib2/communication/Serial/serial_com.cpp
@@ -1,3 +1,5 @@
+#if defined(__linux__)
+
 #include "serial_com.hpp"
 
 SerialCom::SerialCom(std::string path_,
@@ -15,6 +17,7 @@ SerialCom::SerialCom(std::string path_,
     _minChar(minChar_),
     _timeout(timeout_)
 {
+    assert((cflags_ & CSIZE) == 0 && "data size bits belong in eDataPerPacket, not cflags_");
     _fileDesc = open(path_.c_str(), O_RDWR | O_NOCTTY | O_SYNC | O_CLOEXEC);
     if (_fileDesc < 0)
     {
@@ -169,3 +172,5 @@ eState SerialCom::getState() const
 {
     return _state;
 }
+
+#endif
--- a/rover_shared_libs/rover_lib2/src/rover_lib2/communication/Serial/serial_com.cpp
+++ b/rover_shared_libs/rover_lib2/src/rover_lib2/communication/Serial/serial_com.cpp
@@ -1,70 +1,113 @@
 #include "serial_com.hpp"
 
-SerialCom::SerialCom(eBaudRate baudRate_,
-                     eCharSize char_,
-                     bool twoStopBit_,
-                     bool enRead_,
-                     bool ignModem_,
-                     bool parity_,
-                     bool oddParity_,
-                     uint16_t minChar_,
-                     uint16_t timeout_)
+SerialCom::SerialCom(int fileDesc_, eBaudRate baudRate_, eDataPerPacket char_,
+                     tcflag_t cflags_, uint16_t minChar_, uint16_t timeout_)
+                     : _fileDesc(fileDesc_),
+                       _baudRate(std::to_underlying(baudRate_)),
+                       _char(std::to_underlying(char_)),
+                       _cflags(cflags_),
+                       _minChar(minChar_),
+                       _timeout(timeout_),
+                       _state(eState::INACTIVE)
 {
-    _baudRate = static_cast<speed_t>(baudRate_);
-    _char = static_cast<tcflag_t>(char_);
-    _twoStopBit = twoStopBit_;
-    _enRead = enRead_;
-    _ignModem = ignModem_;
-    _parity = parity_;
-    _oddParity = oddParity_;
-    _minChar = minChar_;
-    _timeout = timeout_;
+    if (!this->serialConfig())
+    {
+        LOG_ERROR(Logger::Nodes::SerialCom, "Unable to configure the serial port");
+    }
+    else
+    {
+        this->_state = eState::ACTIVE;
+    }
 }
 
-void SerialCom::serialConfig(int fileDesc_)
+SerialCom::~SerialCom()
+{
+    if (this->_fileDesc >= 0)
+    {
+        close(this->_fileDesc);
+        this->_fileDesc = -1;
+        this->_state = eState::INACTIVE;
+    }
+}
+
+bool SerialCom::serialConfig()
 {
     struct termios tty;
 
     memset(&tty, 0, sizeof(tty));
-    if (tcgetattr(fileDesc_, &tty) != 0)
+    if (tcgetattr(this->_fileDesc, &tty) != 0)
     {
-        std::cout << "tcgetattr failed" << std::endl;
+        std::string errorMsg = "tcgetattr failed: " + std::string(strerror(errno));
+        LOG_ERROR(Logger::Nodes::SerialCom, errorMsg.c_str());
+        this->_state = eState::INACTIVE;
+        return false;
     }
 
-    cfsetospeed(&tty, _baudRate);
-    cfsetispeed(&tty, _baudRate);
+    cfsetospeed(&tty, this->_baudRate);
+    cfsetispeed(&tty, this->_baudRate);
 
-    controlFlagsInit(tty);
+    this->controlFlagsInit(tty);
 
     tty.c_lflag = 0;  // Disables all flags
     tty.c_iflag = 0;  // Disables all flags
     tty.c_oflag = 0;  // Disables all flags
 
-    tty.c_cc[VMIN] = _minChar;
-    tty.c_cc[VTIME] = _timeout;
+    tty.c_cc[VMIN] = this->_minChar;
+    tty.c_cc[VTIME] = this->_timeout;
 
-    if (tcsetattr(fileDesc_, TCSANOW, &tty) != 0)
+    if (tcsetattr(this->_fileDesc, TCSANOW, &tty) != 0)
     {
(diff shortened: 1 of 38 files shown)


rover_shared_libs/rover_lib2/src/rover_lib2/communication/Serial/serial_com.cpp
```diff
@@ -0,0 +1,99 @@
...
+    _twoStopBit = twoStopBit_;
+    _enRead = enRead_;
+    _ignModem = ignModem_;
+    _parity = parity_;
+    _oddParity = oddParity_;
+    _minChar = minChar_;
+    _timeout = timeout_;
+}
+
+void SerialCom::serialConfig(int fileDesc_)
```
emilesavoie: The configuration should all be done in the constructor of the object. There should be no need to call a seperate config method

rover_shared_libs/rover_lib2/src/rover_lib2/communication/Serial/serial_com.cpp
```diff
@@ -0,0 +1,99 @@
+#include "serial_com.hpp"
+
+SerialCom::SerialCom(eBaudRate baudRate_,
```
emilesavoie: This constructor is a bit charged. Its nice to have the ability to control as these features but 95% of applications will use standardized values for all of these. You could use default value parameters to fix this

rover_shared_libs/rover_lib2/src/rover_lib2/communication/Serial/serial_com.cpp
```diff
@@ -0,0 +1,99 @@
...
+                     eCharSize char_,
```
emilesavoie: Initialization of class attributes can be done through initializer list
https://en.cppreference.com/w/cpp/language/initializer_list.html

rover_shared_libs/rover_lib2/src/rover_lib2/communication/Serial/serial_com.cpp
```diff
@@ -0,0 +1,99 @@
...
+                     bool twoStopBit_,
+                     bool enRead_,
+                     bool ignModem_,
+                     bool parity_,
+                     bool oddParity_,
+                     uint16_t minChar_,
+                     uint16_t timeout_)
+{
+    _baudRate = static_cast<speed_t>(baudRate_);
```
emilesavoie: You can use std::to_underlying here:

```suggestion
    _baudRate = std::to_underlying(baudRate_);
```

rover_shared_libs/rover_lib2/src/rover_lib2/communication/Serial/serial_com.cpp
```diff
@@ -0,0 +1,99 @@
...
+    _char = static_cast<tcflag_t>(char_);
```
emilesavoie:
```suggestion
    _char = std::to_underlying(char_);
```

rover_shared_libs/rover_lib2/src/rover_lib2/communication/Serial/serial_com.cpp
```diff
@@ -0,0 +1,99 @@
+#include "serial_com.hpp"
+
+SerialCom::SerialCom(eBaudRate baudRate_,
+                     eCharSize char_,
+                     bool twoStopBit_,
```
emilesavoie: The file descriptor should be an input parameter to the constructor. Ideally, the detection would be dynamic (when your computer detects if a device has been connected). If dynamically detected, we would need to implement a check on whether the port is currently used or not

rover_shared_libs/rover_lib2/src/rover_lib2/communication/Serial/serial_com.hpp
```diff
@@ -0,0 +1,68 @@
...
+    SerialCom(eBaudRate baudRate_,
+              eCharSize char_,
+              bool twoStopBit_,
+              bool enRead_,
+              bool ignModem_,
+              bool parity_,
+              bool oddParity_,
+              uint16_t minChar_,
+              uint16_t timeout_);
+    void serialConfig(int fileDesc_);
```
emilesavoie: This should be a private method called by the constructor

rover_shared_libs/rover_lib2/src/rover_lib2/communication/Serial/serial_com.hpp
```diff
@@ -0,0 +1,68 @@
...
+    void serialWrite(int fileDesc_, const std::string& cmd_);
+    std::string serialRead(int fileDesc_);
+    void controlFlagsInit(termios& tty_);
```
emilesavoie: This should also be a private method

rover_shared_libs/rover_lib2/src/rover_lib2/communication/Serial/serial_com.cpp
```diff
@@ -0,0 +1,99 @@
+#include "serial_com.hpp"
```
emilesavoie, line 3: In general, always refer to class attributes using this keyword

rover_shared_libs/rover_lib2/src/rover_lib2/communication/Serial/serial_com.cpp
```diff
@@ -0,0 +1,99 @@
...
+    _ignModem = ignModem_;
+    _parity = parity_;
+    _oddParity = oddParity_;
+    _minChar = minChar_;
+    _timeout = timeout_;
+}
+
+void SerialCom::serialConfig(int fileDesc_)
+{
+    struct termios tty;
```
emilesavoie: This should be a class attribute

rover_shared_libs/rover_lib2/src/rover_lib2/communication/Serial/serial_com.hpp
```diff
@@ -0,0 +1,68 @@
...
+
+    speed_t _baudRate;
```
emilesavoie: All attributes should private in this case. If the user needs to access them, use setters and getters

rover_shared_libs/rover_lib2/src/rover_lib2/communication/Serial/serial_com.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+void SerialCom::serialWrite(int fileDesc_, const std::string& cmd_)
+{
+    ssize_t bytesWritten = write(fileDesc_, cmd_.c_str(), cmd_.size());
+    (void)bytesWritten;
+}
+
+std::string SerialCom::serialRead(int fileDesc_)
+{
+    char buffer[256];
```
emilesavoie: Should this be a constant? If so, define as static const in class definition

rover_shared_libs/rover_lib2/src/rover_lib2/communication/Serial/serial_com.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+    if (n > 0)
+    {
+        response.assign(buffer, n);
+    }
+
+    return response;
+}
+
+void SerialCom::controlFlagsInit(termios& tty_)
```
emilesavoie: The user could directly imput a cflag instead of 5 different booleans especially since there is a standard cflag from what I have read (this could be declared as a default parameter to the class). If needed, multiple flags could then be added to form the desired masl

rover_shared_libs/rover_lib2/src/rover_lib2/communication/Serial/serial_com.cpp
```diff
@@ -0,0 +1,99 @@
...
+    tty.c_lflag = 0;  // Disables all flags
+    tty.c_iflag = 0;  // Disables all flags
+    tty.c_oflag = 0;  // Disables all flags
+
+    tty.c_cc[VMIN] = _minChar;
+    tty.c_cc[VTIME] = _timeout;
+
+    if (tcsetattr(fileDesc_, TCSANOW, &tty) != 0)
+    {
+        std::cout << "tcsetattr failed" << std::endl;
```
emilesavoie: Error logging is good, but there are a couple things here.
1. Use our custom logger for all logs since this allows logs to propagate to multiple processes
2. If an error is encountered, should there only be a log or an action be taken place? Depends what this error checks for exactly

rover_shared_libs/rover_lib2/src/rover_lib2/communication/Serial/serial_com.hpp
```diff
@@ -0,0 +1,68 @@
...
+    B_1152000 = B1152000,
+    B_1500000 = B1500000,
+    B_2000000 = B2000000,
+    B_2500000 = B2500000,
+    B_3000000 = B3000000,
+    B_3500000 = B3500000,
+    B_4000000 = B4000000
+};
+
+enum class eCharSize : tcflag_t
```
emilesavoie: This naming is confusing here. Both the name of the enum as well as the C_S5, C_S6, not very intuitive

emilesavoie (changes requested):

emilesavoie (changes requested):
There are quite a few comments here so I will let you start on this. The basic functionnality is what we want but the structure lacks a bit of finesse

rover_shared_libs/rover_lib2/src/rover_lib2/communication/Serial/serial_com.cpp
JoeyBilodeauL: Didnt make it dynamic. If needed, could add in the future

rover_shared_libs/rover_lib2/src/rover_lib2/communication/Serial/serial_com.cpp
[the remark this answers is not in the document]
JoeyBilodeauL: Added a reconnect function. I had to include all member variables originally included ( _baudRate, _char, _cflags, etc.) Also, possible to include the cooldown for retries on the caller side instead of being built in to add more versatility?

PR #291, opened 2026-03-11
Fixing lint script + linting

AraaRobot:


.clang-format
```diff
@@ -59,7 +59,7 @@ SpacesInCStyleCastParentheses: 'false'
 SpacesInContainerLiterals: 'false'
 SpacesInParentheses: 'false'
 SpacesInSquareBrackets: 'false'
-Standard: Cpp23
+Standard: Latest
```
emilesavoie: Lets keep a specific versiono here to avoid unexpected behaviour. We`ll move on to the next version whenever appropriate

emilesavoie (changes requested):

.clang-format
AraaRobot: Should i use the corresponding version then? Latest is currently the equivalent of Cpp20. Or should we go back to Cpp11?

emilesavoie: Cpp20 is okay

emilesavoie (approved):

PR #300, opened 2026-06-05
Fixing waypoints from JS --> Qt

AraaRobot:
While fixing waypoints, found a second problem. PopUps were not returning until app was closed.

Explanation by Claude:

# What broke moving from Qt5 on Ubuntu 22.04 to Qt6 on Ubuntu 24.04
## Problem 1 — waypointCreated slot never firing

```cpp
void waypointCreated(const QString& name_, double latitude_, double longitude_, QString& id_);
void waypointCreated(const QString& name_, double latitude_, double longitude_, const QString& id_);
```

The slot had a QString& (non-const reference) parameter. QWebChannel works by deserializing JSON into temporary values and injecting them into slot calls — there's no persistent object to hand a reference to. Qt5 let this slip through, Qt6 strictly validates that all slot parameters are serializable types and silently rejects the slot if they aren't. Fixed by changing QString& to const QString&, which is valid because C++ allows const references to bind to temporaries with extended lifetime.

## Problem 2 — sendQuestionPopUp blocking forever
The old code used QMetaObject::invokeMethod with a nested QEventLoop to simulate blocking. In Qt5, invokeMethod with a lambda defaulted to Qt::DirectConnection when called from the main thread, so the lambda ran synchronously before waitForAnswerLoop.exec() was even reached. In Qt6, lambda overloads default to Qt::QueuedConnection, so the lambda gets posted to the event queue instead — the QEventLoop starts spinning, the dialog opens fine, but the quit signal and the app's aboutToQuit interact in a way that breaks the unblocking, causing it to hang until app shutdown.
Fixed by ditching the nested QEventLoop entirely and replacing it with a direct call check plus Qt::BlockingQueuedConnection for cross-thread cases:

If already on the main thread → call QMessageBox::question directly
If on a worker thread → use Qt::BlockingQueuedConnection which blocks the calling thread and posts to the main thread's event loop, then resumes the caller when the dialog returns

This matters especially for QSshWorker, where the popup is called from background worker tasks that genuinely need the user's answer before deciding whether to continue the file transfer.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -59,7 +59,7 @@ SpacesInCStyleCastParentheses: 'false'
 SpacesInContainerLiterals: 'false'
 SpacesInParentheses: 'false'
 SpacesInSquareBrackets: 'false'
-Standard: Cpp20
+Standard: c++20
 TabWidth: '2'
 UseTab: Never
 SortIncludes: 'false'
--- a/rover_gui/src/Global/Constant/other/JS/bridge.js
+++ b/rover_gui/src/Global/Constant/other/JS/bridge.js
@@ -32,57 +32,62 @@ class Bridge
             const qtBridge = channel.objects.bridge;
             window.qtBridge = qtBridge;
 
-            const self = this;
-
-            qtBridge.clearPath.connect(function () 
+            qtBridge.clearPath.connect(() =>
             {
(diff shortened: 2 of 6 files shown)


emilesavoie (approved):
Looking good!

